Issue #63, opened 2018-05-03
Travis keeps failing for jlabce 1.2


wdconinc:
Related to AnyMethod for G4GenericMessenger. Probably due to `const G4String&` which used to be buggy... Need to test on ifarm or docker with `docker build -t jeffersonlab/remoll .` (sudo required).

```
In file included from /usr/include/Geant4/G4GenericMessenger.hh:36:0,
                 from /home/travis/build/JeffersonLab/remoll/src/remollSteppingAction.cc:7:
/usr/include/Geant4/G4AnyMethod.hh: In instantiation of â€˜void G4AnyMethod::FuncRef1<S, T, A0>::operator()(void*, const string&) [with S = void; T = remollSteppingAction; A0 = const G4String&; std::string = std::basic_string<char>]â€™:
/home/travis/build/JeffersonLab/remoll/src/remollSteppingAction.cc:88:1:   required from here
/usr/include/Geant4/G4AnyMethod.hh:185:12: error: cannot bind â€˜std::basic_istream<char>â€™ lvalue to â€˜std::basic_istream<char>&&â€™
       strs >> a0;
            ^
In file included from /usr/include/c++/4.8/sstream:38:0,
                 from /usr/include/c++/4.8/complex:45,
                 from /usr/include/Geant4/G4Types.hh:67,
                 from /usr/include/Geant4/G4ios.hh:39,
                 from /usr/include/Geant4/globals.hh:49,
                 from /home/travis/build/JeffersonLab/remoll/include/remollSteppingAction.hh:6,
                 from /home/travis/build/JeffersonLab/remoll/src/remollSteppingAction.cc:1:
/usr/include/c++/4.8/istream:872:5: error:   initializing argument 1 of â€˜std::basic_istream<_CharT, _Traits>& std::operator>>(std::basic_istream<_CharT, _Traits>&&, _Tp&) [with _CharT = char; _Traits = std::char_traits<char>; _Tp = const G4String]â€™
     operator>>(basic_istream<_CharT, _Traits>&& __is, _Tp& __x)
     ^
make[2]: *** [CMakeFiles/remoll-lib.dir/src/remollSteppingAction.cc.o] Error 1
make[1]: *** [CMakeFiles/remoll-lib.dir/all] Error 2
make: *** [all] Error 2
```

wdconinc:
Fixed in 9e0ad05

G4GenericMessenger to method with const G4String call-by-reference arg fails in gcc < 5.0
    
[...]

PR #65, opened 2018-05-03
Integration pion

wdconinc:


wdconinc:
Travis will fail per #63

PR #66, opened 2018-05-11
Integration pion into develop

wdconinc:


wdconinc:
Some of these validation errors are the GDML schema running behind the actual Geant4 implementation and support (e.g. the one referring to attribute phi (replication in phi direction) or to property (optical properties in GDML)). The one with duplicate ID values is because of tricks I play with GDML...

In any case, I should fix these since they can be fixed.

The main issue is (continues to be) the unexplained stuck tracks. Normally these are a sign of geometry overlap, and I suspect the same is true here due to the tiny bit of overlap between the 7 sectors that fill a full azimuthal cylinder. I will look into this.

Reopening as new issue. Thanks for reporting.

PR #67, opened 2018-05-11
Develop

wdconinc:


cameronc137:
For this commit, it appears to be adding some methods for calculating mrem and NEIL radiation damage for various particles and energy ranges. Is there a concise resource for reading up on NEIL and mrem conversions? Also, would these calculated values get stored into the output trees alongside the other original variables?

cipriangal:
Documentation:
https://prex.jlab.org/cgi-bin/DocDB/private/ShowDocument?docid=85

The class was intended to be used as a post simulation radiation analysis. While adding this information to the output could be somewhat useful I am afraid it will just add additional variables that in the vast majority of cases will not used for much. What I did in the prex simulation is to create a different tree for particular detector hits (https://github.com/cipriangal/prexSim/blob/master/reduceTree.cc) 

We can discuss and if people feel this is well suited to be placed in the output we can easily implement it.

PR #68, opened 2018-05-12
Release v2.0.0

wdconinc:


wdconinc:
@cameronc137  v2.0.0 will not have split out geometry yet, but as soon as v2.0.0 is released I'll move develop to submodule geometry

geometry_sculpt/pionDetectorSystem.gdml
```diff
@@ -0,0 +1,130 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cameronc137: The pion geometry was added to the /geometry_sculpt/ folder, but the geometry with the up to date shielding and collimator designs is in the /geometry/ folder

analysis/remoll.C
```diff
@@ -1,51 +0,0 @@
-#define remoll_cxx
```
cameronc137: It seems that some of these analysis scripts are machine/path or output file specific. This is fine, but it may be nice to have some more ownership, history or context documentation, or make them more generic. 

Also, are people encouraged to commit their analysis scripts they use, even if they are just for specific day to day tasks, not general consumption?

include/remollGenericDetector.hh
```diff
@@ -23,10 +24,37 @@
...
+    private:
+        static G4GenericMessenger* fStaticMessenger;
+        static std::set<remollGenericDetector*> fGenericDetectors;
+        static void InsertGenericDetector(remollGenericDetector* det) {
+          fGenericDetectors.insert(det);
+        }
+        static void EraseGenericDetector(remollGenericDetector* det) {
+          fGenericDetectors.erase(det);
+        }
+        void SetAllEnabled() {
```
cameronc137: I remember these macro commands were being discussed recently, and in general it would be nice to have a catalog of macro commands, what they are good for, and where in the code they are documented (now that things are decentralized away from the master messenger class except for the seed control).

include/remollGenericDetector.hh
```diff
@@ -35,6 +63,34 @@ class remollGenericDetector : public G4VSensitiveDetector {
 	virtual G4bool ProcessHits(G4Step*,G4TouchableHistory*);
 	virtual void EndOfEvent(G4HCofThisEvent*);
 
+	virtual void SetDetectorType(G4String det_type) {
```
cameronc137: What are these new detector types, and are they physically relevant or just ideal detectors for monitoring specific kinds of flux?

include/remollGenericDetectorHit.hh
```diff
@@ -26,7 +28,11 @@ class remollGenericDetectorHit : public G4VHit {
 
 	// Position and momentum in lab coordinates
 	G4ThreeVector f3X;
+	G4ThreeVector f3Xl;
```
cameronc137: What are these new "l" vectors? Their descriptions and new names should be added to the variables.txt file as well.

analysis/tests/test_general.C
```diff
@@ -21,14 +21,22 @@ void test_general (const TString& inputdir, const TString& inputname)
   c1.SaveAs(outputdir + "/" + inputname + "_" + "h_hit_pid.png");
   T.Draw("hit.det >> h_hit_det(4500,0,4500)","",hopt);
   c1.SaveAs(outputdir + "/" + inputname + "_" + "h_hit_det.png");
-  T.Draw("hit.r   >> h_hit_r(150,0.0,1.5)","hit.det==28",hopt);
+  T.Draw("hit.r/m >> h_hit_r(150,0.0,1.5)","hit.det==28",hopt);
```
cameronc137: These hit.XX variables should be changed to part.XX (as well as in other analysis scripts)

include/remollGenericDetectorHit.hh
```diff
@@ -36,6 +42,38 @@ class remollGenericDetectorHit : public G4VHit {
 	// Process generator type
 	G4int    fGen;
 
+    public:
+      const remollGenericDetectorHit_t GetGenericDetectorHitIO() const {
+        remollGenericDetectorHit_t hit;
```
cameronc137: Does this hit.XX variable need to be updated to match the part.XX naming convention in the new multi-thread safe IO, or is it fine as it is? I would assume it's OK as it is.

include/remollParallelWorldConstruction.hh
```diff
@@ -0,0 +1,34 @@
...
+
+class G4LogicalVolume;
+class G4VPhysicalVolume;
+class G4GDMLParser;
+
+//
+/// A parallel world construction class
+///
+/// - void Construct()
+///     creates a parallel world in the mass world and parameterized volumes
```
cameronc137: What is a parallel world?

src/remollGenExternal.cc
```diff
@@ -0,0 +1,132 @@
...
+
+remollGenExternal::remollGenExternal()
+: remollVEventGen("external"),
+  fFile(0), fTree(0),
+  fEntry(0), fEntries(0),
+  fEvent(0), fHit(0),
+  fDetectorID(0)
+{
+  // Add to generic messenger
+  fThisGenMessenger->DeclareMethod("file",&remollGenExternal::SetGenExternalFile,"External generator event filename");
```
cameronc137: What is the macro command exactly for utilizing external files for generating data, and what kind of a file and what structure is required to do this? Does it sample the exact entries in the data file or does it do some form of interpolation?

src/remollIO.cc
```diff
@@ -80,77 +76,24 @@ void remollIO::InitializeTree()
...
-    fTree->Branch("hit.vy",   &fGenDetHit_Vy,   "hit.vy[hit.n]/D");
-    fTree->Branch("hit.vz",   &fGenDetHit_Vz,   "hit.vz[hit.n]/D");
-
-    fTree->Branch("hit.p",    &fGenDetHit_P,   "hit.p[hit.n]/D");
-    fTree->Branch("hit.e",    &fGenDetHit_E,   "hit.e[hit.n]/D");
-    fTree->Branch("hit.m",    &fGenDetHit_M,   "hit.m[hit.n]/D");
-
-    fTree->Branch("hit.colCut",    &fCollCut,     "hit.colCut/I");
-
+    fTree->Branch("hit",      &fGenDetHit);
```
cameronc137: Is this "hit" different than the previous "hit" that was renamed as "part"?

geometry_sculpt/pionDetectorSystem.gdml
wdconinc: Yeah, that's next... As long as people keep copying geometry to new folders it's not making it any easier to merge.

analysis/remoll.C
wdconinc: Commit all you can, even if it's under analysis/cameron/. That way nothing is lost when you graduate. Don't commit large binary blobs though.

src/remollSteppingAction.cc
```diff
@@ -9,30 +9,59 @@
 remollSteppingAction::remollSteppingAction()
 : fDrawFlag(false),fEnableKryptonite(true)
 {
+  AddKryptoniteCandidate("VacuumKryptonite");
```
cameronc137: These material names assume that a corresponding material is defined in the materials.xml file correct? Should we go ahead and make the transition over to G4_XXXX variable names to prevent issues with over defined (in sub gdml files or alternate extra materials.xml files) material definitions?

include/remollGenericDetector.hh
wdconinc: Working on that. Geant4 can write a html help directory with all possible commands and their syntax. Unfortunately you can't specify the path where to write it...

include/remollGenericDetectorHit.hh
wdconinc: Local geometry, in the frame attached to the detector solid instead of global frame.

analysis/tests/test_general.C
wdconinc: Nope, part is for the particles that are generated. hit is for those that are detector.

include/remollParallelWorldConstruction.hh
wdconinc: Parallel world can have SDs that are intersecting different volumes. The goal is to put the virtual detector planes (like the flat disk det = 28 for main detector array) in the parallel world so it's still available but doesn't mess with the actual material geometry. Support only added in geant4.10

src/remollVertex.cc
```diff
@@ -6,8 +6,8 @@
 remollVertex::remollVertex(){
     // Some default material
     fMaterial = NULL;
-    fBeamE = 0.0*GeV;
-    fRadLen = 0.0;
+    fBeamEnergy = 0.0*GeV;
+    fRadiationLength = 0.0;
```
cameronc137: Just to be precise, does this radiation length need some default length units, similar to the beam energy assuming GeV?

src/remollGenExternal.cc
wdconinc: The input file format for gen external is the output file format :-) Use case: you are only interested in what happens to electrons after the big lead wall downstream before the pion system. You can simulated lots of events to get the few electron showers that pass through the lead wall. Those can all be stored in a remollout.root file thanks to a virtual detector right behind the lead wall (potentially even set to kryptonite). Next step, use that file of events as input: events will be generated at the pion wall instead of main target, so you have faster simulations of the area downstream.

src/remollIO.cc
wdconinc: hit was not renamed as part. Part (sic) of ev was renamed as part.

analysis/remoll.C
cameronc137: Ok. I have a little repo going on the sbujlab organization that I am storing various applied versions of similar code on. Would this be something worth submoduling and tracking, or just copy over those into analysis sub folders in here as a storage mechanism?

include/remollGenericDetectorHit.hh
cameronc137: This sounds pretty useful, great! One concern though, in general, is if there are too many variables then the output files could get quite large. Is there a technique (other than preprocessor directive flagged ifdef statements) for trimming out variables or entire segments of simulation for specific applications of the main code?

src/remollGenExternal.cc
cameronc137: Ok, interesting. I see there are macro commands for picking the data file and which detector to use. Is the macro command just /remoll/file and that does it, or is there another specifier too?

cameronc137: Also, would there be a possibility to add an option for generating the new event's vertices some distance z_offset upstream of the positions stored in the file? Backtracking or something, to allow, for instance, for scintillation and cherenkov optical physics to accumulate as the particle streams towards the detector it had hit? Or would it be better to just place an ideal vacuum detector upstream of the one we want to simulate and use that as the upstream z_offset for simplicity?

src/radDamage.cc
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  return interpolatedValue;
+}
+
+double radDamage::interpolate(vector<double> xV, vector<double> yV,double energy){
+  if( energy <= xV.front() ) return 0;
+  else if( energy >= xV.back() ) return yV.back();
+
+  int lowIndex  = int(lower_bound(xV.begin(),xV.end(),energy) - xV.begin() ) - 1;
```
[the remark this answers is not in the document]
cipriangal: Compiled for me on both my mac and my linux box. What is your environment?

include/remollDetectorConstruction.hh
```diff
@@ -50,12 +55,16 @@ class remollDetectorConstruction : public G4VUserDetectorConstruction
 
     void DumpElements();
     void DumpMaterials();
-    void DumpGeometricalTreeFromWorld(G4bool surfchk = false) {
+    void DumpGeometry(G4bool surfchk = false) {
```
cameronc137: Is this the /remoll/dumpgrometry true command that is used in the runexample.mac macro?

If so how does the DumpGeometricalTree function work.

Also - I'm assuming dump here means erase, not beamdump?

macros/runexample.mac
```diff
@@ -45,18 +45,20 @@
...
 
-#Enable optical photons and processes
-#/remoll/optical true
-
 # Make interactions with W, Cu, and Pb
 # realistic rather than pure absorbers
-/remoll/kryptonite true
+/remoll/kryptonite/set true
 
 /process/list
```
cameronc137: I'm not sure what this process/list is, but for some time now, running any beam on non-trivial nuclei/shielding/collimators will print dozens of messages like this to the screen, so I think there is some issue with looking up scattering cross sections:

"NeutronHP: /Elastic file for Z = 6, A = 12 is not found and NeutronHP will use /share/apps/geant4/geant4.10.01.p03-install/share/Geant4-10.1.3/data/G4NDL4.5/Elastic/CrossSection/6_nat_Carbon"

include/remollGenericDetectorHit.hh
```diff
@@ -36,6 +42,38 @@ class remollGenericDetectorHit : public G4VHit {
...
+        hit.z  = f3X.z();
+        hit.xl = f3Xl.x();
+        hit.yl = f3Xl.y();
+        hit.zl = f3Xl.z();
+        hit.r  = sqrt(f3X.x()*f3X.x()+f3X.y()*f3X.y());
+        hit.ph = f3X.phi();
+        hit.px  = f3P.x();
+        hit.py  = f3P.y();
+        hit.pz  = f3P.z();
+        hit.pxl = f3Pl.x();
```
cameronc137: In root output files these hit.p_l values are all -1e+09, which seems to be a placeholder null value.

src/remollBeamTarget.cc
```diff
@@ -420,42 +447,43 @@ remollVertex remollBeamTarget::SampleVertex(SampType_t samp)
...
     // Sample beam energy based on radiation
     // We do this so it doesn't affect the weighting
     //
     // This can be ignored and done in a generator by itself
 
-    G4double  Ekin = fBeamE - electron_mass_c2;
-    G4double  bt   = fRadLen*4.0/3.0;
-    G4double  prob_sample, eloss, sample, env, value, ref;
+    G4double  Ekin = fBeamEnergy - electron_mass_c2;
+    G4double  bt   = fRadiationLength * 4.0 / 3.0;
```
cameronc137: What is this bt? And if it depends on fRadiationLength, does it need to have units (currently unitless)

include/remollGenericDetectorHit.hh
wdconinc: FYI Qweak ROOT trees have 1170 fields... so I think we're fine :-) but I do get your concern. Will think about this.

src/radDamage.cc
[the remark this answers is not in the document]
wdconinc: "GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard"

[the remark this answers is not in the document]
wdconinc: If you have a compiler error, we can try to work around until you can upgrade to > CentOS 7.

macros/dumpgeometry.mac
```diff
@@ -0,0 +1,7 @@
+# This must be called before initialize
+/remoll/setgeofile geometry_sculpt/mollerMother.gdml
+
+# This must be explicitly called
+/run/initialize
```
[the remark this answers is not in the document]
wdconinc: Initialize creates detector construction. Therefore, setting gdml file (to non default) is required before  initialize. Moving to single geometry directory in submodule may mean we can initialize automatically.

include/remollDetectorConstruction.hh
wdconinc: Dump == Print. Bad name. It's recursively printing the geometry tree. With parallel worlds this requires specifying which world.

macros/runexample.mac
wdconinc: Most of those are due to metastable isotopes that are in the natural abundances of G4_ NIST materials. There's a way to hide these warnings, but I haven't committed that (yet).

include/remollGenericDetectorHit.hh
wdconinc: Already removed in feature-spin-precession

wdconinc: These were "local momentum", but that doesn't make sense (for stationary detectors).

analysis/remoll.C
wdconinc: Either way. Don't get too crazy on submodules since they don't recurse by default. I'd say keep it in the main repository to the extent it would likely be useful to others.

vis/vis.mac
```diff
@@ -52,8 +52,8 @@
 /vis/viewer/zoom 5.0
 /vis/viewer/set/lightsThetaPhi 40 50
 #New cutaway plane setting
-/vis/viewer/set/cutawayMode intersection
-/vis/viewer/addCutawayPlane 0.0 0.0 0.0 m 1.0 0.0 0.0
+#/vis/viewer/set/cutawayMode intersection
```
cameronc137: Personally I favor the cutaway as default functionality in the vis macro, particularly when the full hall walls are included.

cameronc137:
Other than a few small questions, a bit of further documentation in places, and the GCC error on the radiation class, this version of the simulation works just fine and runs for me on the SBU cluster as desired.

cipriangal:
When running I get:
Warning: No target mother volume found.
which is produced by https://github.com/JeffersonLab/remoll/blob/release-v2.0.0/src/remollBeamTarget.cc#L124
Don't remember this being an issue before.

src/radDamage.cc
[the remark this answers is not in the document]
cameronc137: The reason is that it is prohibitively difficult to upgrade the software on our cluster, so backwards compatible workarounds would be appreciated for the time being.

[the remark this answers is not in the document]
cipriangal: I'll add this to my todo list. @wdconinc is there a compiler flag we can use to check cxx version?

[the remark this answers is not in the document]
wdconinc: It's in CMakeLists. We could add a line for C++11 support not found like this:
```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNO_COMPILER_SUPPORTS_CXX11")
```

wdconinc:
I now see what the target error is that @cipriangal is getting :-)  It only happens when you use geometry/, not when you use geometry_sculpt/, so I wasn't able to reproduce at first. I'll fix the geometry/ error (need to add aluminum target windows). Can you confirm that this is indeed when using geometry/?

wdconinc:
But I will only update geometry/targetDaughter.gdml to work, for now.

geometry_sculpt/pionDetectorSystem.gdml
wdconinc: Fixed in develop 08e27295fd83c17f98cbfae8847e3387c09a6725

include/remollGenericDetectorHit.hh
wdconinc: Fixed in develop 757cb11ba7cfbb12f6cc85bfb2b2e9bec44e3d34

analysis/tests/test_general.C
wdconinc: NOP

include/remollParallelWorldConstruction.hh
wdconinc: NOP

src/remollIO.cc
wdconinc: NOP

src/remollSteppingAction.cc
wdconinc: Moved to #70

src/remollVertex.cc
wdconinc: Units are in internal G4 units, as given by actual_position_in_volume/material->GetRadlen() in e.g. https://github.com/JeffersonLab/remoll/blob/fd7d856136ece8ed22a85c4cfc1145bb20e4cb69/src/remollBeamTarget.cc#L325. Also zero is zero in any units.

NOP

src/radDamage.cc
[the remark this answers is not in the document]
wdconinc: Looking at this some more, why does it even try to compile radDamage.cc? It's not linked anywhere.

For now, removing radDamage from list of sources to compile until bypassed or systems upgraded to full C++11 support.

Fixed in develop 642d80a

macros/dumpgeometry.mac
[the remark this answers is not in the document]
wdconinc: Default geometry changed to geometry/

NOP

cipriangal:
@wdconinc: for the target not found Warnings/Errors I am just running macros/runexample with the release-v2.0.0 branch. The geometry file it uses is 
/remoll/setgeofile geometry_sculpt/mollerMother.gdml

The first warning comes after generating the kryptonite materials (really at the top of the output). While the set of other warnings come after the GDML parsing and before some **det_XXX is not found in /remoll/** warnings/errors.

src/radDamage.cc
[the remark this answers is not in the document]
cipriangal: @cameronc137 can you see if this will work for you. If not I'll just modify the class.

include/remollDetectorConstruction.hh
wdconinc: Fixed in develop ab6623c

src/remollBeamTarget.cc
wdconinc: Interesting... fRadiationLength is in units of length/material->GetRadlen(), so bt is too, and then bt is used as an exponent. But compare:
https://github.com/JeffersonLab/remoll/blob/fd7d856136ece8ed22a85c4cfc1145bb20e4cb69/src/remollBeamTarget.cc#L325
https://github.com/JeffersonLab/remoll/blob/fd7d856136ece8ed22a85c4cfc1145bb20e4cb69/src/remollBeamTarget.cc#L328
It seems that cumulative_radiation_length has either no unit, or it has unit 1/density. This needs further investigation.

wdconinc: Reported as new issue #71

vis/vis.mac
wdconinc: Fixed in develop 8450f74

src/radDamage.cc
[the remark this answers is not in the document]
cameronc137: Adding the CMakeLists extra flag to the list of flags doesn't fix it, but #642d80a does fix it.

wdconinc:
"Warning: No target mother volume found" as reported by @cipriangal above is now fixed.

Fixed in develop 71e4103

include/remollGenericDetector.hh
wdconinc: Fixed in #73

src/remollBeamTarget.cc
wdconinc: Likely not an issue, see #71.

src/remollVertex.cc
wdconinc: Discussed in issue #71.

src/remollGenExternal.cc
wdconinc: NOP

include/remollGenericDetector.hh
wdconinc: The goal is to allow some differentiation of sensitive detectors as defined in GDML, in particular when we turn on optical photons.

Use case:
> I would like to simulate the response of the main detector array, incl both PMT photocathode and quartz itself, when optical photons are turned on.
> I care about simulating optical photons and want the optical photons to generate a hit when they are detected in the photocathode sensitive detector.
> But I don't care about getting a hit entry for every single optical photon bouncing in the quartz, at every step (trust me, you don't want to fill up each event with 10k+ hits like that, it's ugly). I am however interested in storing the hits generated by the electrons as they go through the quartz.
> So, how do I tell the quartz to only register hits by a subset of tracks? SetDetectorType charged will only generate hits for charged particles.

There must be a better way to solve this than by defining (currently) three types of detectors:
- charged
- lowenergyneutral
- opticalphoton

Suggestions for better solution welcome. I want something that works well as default (i.e. behaves sensibly when nothing is specified, likely registering all hits except for optical photons), but where that default can be modified with aux tags in GDML.

wdconinc: Since this was mainly a request for clarification, no action needed.

NOP

macros/runexample.mac
wdconinc: These warnings can be suppressed with
```
G4ParticleHPManager::GetInstance()->SetVerboseLevel(level);
```
somewhere at the end of `remollPhysicsList::SetVerboseLevel`.

However, the G4ParticleHPManager does not want to be shut up silently and complains with:
```
You are setting a new verbose level for Particle HP package.
the new value will be used in whole of the Particle HP package, i.e., models and cross sections for Capture, Elastic, Fission and Inelastic interaction.
```

So, pick your poison: a long list of warnings that arguably convey information or another warning. For now I will set verbose level to zero and warn our users.

Remaining warnings that can't be hidden:
```
@@@ G4ParticleHPInelastic instantiated for particle neutron data directory variable is G4NEUTRONHPDATA pointing to /usr/local/geant4/pro/share/Geant4-10.4.1/data/G4NDL4.5/Inelastic
@@@ G4ParticleHPInelasticData instantiated for particle neutron data directory variable is G4NEUTRONHPDATA pointing to /usr/local/geant4/pro/share/Geant4-10.4.1/data/G4NDL4.5
@@@ G4ParticleHPInelastic instantiated for particle neutron data directory variable is G4NEUTRONHPDATA pointing to /usr/local/geant4/pro/share/Geant4-10.4.1/data/G4NDL4.5/Inelastic
```

Fixed in develop 0ff7db3.

macros/runexample.mac
```diff
@@ -28,8 +30,6 @@
 #/remoll/beam_dth 0 mrad
 #/remoll/beam_dph 0 mrad
 
-/remoll/beamene 11 GeV
-
 #/remoll/gen beam
 /remoll/gen moller
```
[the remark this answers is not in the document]
wdconinc: Remaining `/remoll/gen` warning fixed in develop d2aeb61

wdconinc:
So @cipriangal @cameronc137 I think I've fixed most of the issues you caught (thanks!), or at least pushed them into dedicated issues if I didn't consider them blockers (#68 #70 #71). Can you give it another spin? Then I can merge it into master and release the new major 2.0.0 version.

cameronc137:
Running the simulation with the updated target sampling aux type works as intended, though I am getting mountains of G4 validation errors (but it continues to work and runs as intended) as it reads through the GDML. I am just running the same geometry I had used before, but with the target sampling stuff updated to match commit fd7d856

A new problem (I mentioned somewhere else) is the need to move the root output file away from the rootlogon.c file in order for it to load properly. This is a new problem that isn't really important, but I don't fully understand why it is happening.

I also get segfaults at the end of the run when everything starts shutting down - this has been the case for some time now, so it is not a new issue.

The G4 validation errors look like this:

G4GDML: VALIDATION ERROR! attribute 'auxtype' is not declared for element 'auxiliary' at line: 1480
G4GDML: VALIDATION ERROR! attribute 'auxvalue' is not declared for element 'auxiliary' at line: 1480
G4GDML: VALIDATION ERROR! no declaration found for element 'auxiliary' at line: 1481
G4GDML: VALIDATION ERROR! attribute 'auxtype' is not declared for element 'auxiliary' at line: 1481

wdconinc:
About the validation errors (in particular those with materials):
1. if you don't define each material you use in the GDML file itself (possibly through use of ENTITY), then you will get a GDML validation error.
2. if you do define each material you use in the GDML file itself (possibly through use of ENTITY), then you will get a duplicate G4Material error.
3. if you use G4_XXX materials, then you will get a GDML validation error per 1. above.
4. if you use G4_XXX materials, then you will not get optical properties unless you override the material and get duplicate G4Material errors per 2. above.

wdconinc:
Ultimately GDML validation errors (wrt materials) are just that: failure to validate the GDML in the (narrow) context of individual GDML files. It is true that the individual GDML files are not valid, because they are part of a larger system of files. As long as the full set of GDML files satisfy the validation, then that is sufficient.

wdconinc:
Ok, so the issues left:
1. rootlogon.C not working (can you post the crash you get, @cameronc137 ?). My suggestion is to use build/reroot for anything that requires access to remollX objects. For merely using the tree, use regular root and use the reflection provided by root itself.
2. Validation errors. Added information but wondering if there's more to this.
3. Segfault at closing. Seems to be the case when closing before a run has been started/completed. Not sure what the reason is, and neither valgrind or gdb are giving useful stack traces. Since not affecting actual runs, I'm hoping we can forgive this one, file as issue, and move on with the release.

cameronc137:
For 1. the error messages are

`dlopen error: /share/apps/geant4/geant4.10.01.p03-install/lib64/libG4particles.so: cannot allocate memory in static TLS block
Load Error: Failed to load Dynamic link library /home/cameronc/gitdir/release-v2.0.0/build/./libremoll.so
Warning: Return value of void rootlogon() ignored rootlogon.C:4:
root [0] 
Attaching file remollout.root as _file0...
dlopen error: /share/apps/geant4/geant4.10.01.p03-install/lib64/libG4particles.so: cannot allocate memory in static TLS block
Load Error: Failed to load Dynamic link library /share/apps/root/lib/libThread.so
Error in <TCint::AutoLoadCallback>: failure loading dependent library libThread.so for class TFile
dlopen error: /share/apps/geant4/geant4.10.01.p03-install/lib64/libG4particles.so: cannot allocate memory in static TLS block
Load Error: Failed to load Dynamic link library /share/apps/root/lib/libRIO.so
Error in <TCint::AutoLoadCallback>: failure loading library libRIO.so for class TFile
Error: Function Open("remollout.root") is not defined in current scope  (tmpfile):1:
*** Interpreter error recovered ***`

cameronc137:
For 2. the problem arises from copying the geometry folder from the main directory into the build directory (which is my usual tactic for getting the geometry to sit next to the built executable and the macro folder, but easily adjusted) - this fails to copy the symlinked schema folder correctly and so it gives validation errors for most of the features of GDML - interestingly it still manages to actually simulate stuff, though I'm not sure if it is in fact treating the volumes and materials correctly with the bad schema (maybe the correct schema folder gets found and loads later on, but at this early stage gives all the validation error messages).

wdconinc:
Schema is just the grammar that the file has to satisfy. Geant4 provides all functionality independent of this grammar (of course, it still has to look like valid xml for geant4 to be able to parse it).

wdconinc:
I think we should get rid of rootlogon and use instead one of the following two options, which is more robust:
1. LD_PRELOAD=build/libremoll.so root
2. build/reroot

Both have access to internal remoll objects, as demonstrated by
```
remollEvent_t ev;
```

PR #83, opened 2018-05-25
Merge cadmesh functionality, only compile in if its available...

wdconinc:
Mainly creating the PR to provide a path. More functionality needed in branch before we can do that:
- rebase branch against current develop
- check if cadmesh is available, continue gracefully if not
- compile cadmesh functionality only if available
- messenger to select cadmesh input


wdconinc:
Rebased against develop (only conflicts in CMakeLists.txt due to changed targets and in macro/vis.mac due to removed in develop and modified in branch)

Modified CMakeLists.txt to require
```
cmake -DUSE_CADMESH=1 -Dcadmesh_DIR=<path> ..
```

wdconinc:
Compile cadmesh functionality only if available as of cadmesh 7764876.

wdconinc:
Branch is missing functionality.

I do think there's value in the approach of using cadmesh for **select** volumes. Since cadmesh only defines solids, not material assignment into a logical volume or placement as a physical volume, we need additional meta data.

That metadata is basically what gdml provides. Suggested solution: aux field that would allow the following in gdml:
```
<solids>
  <box lunit="mm" name="bounding_box" x="4000" y="4000" z="4000"/>
  <tube aunit="deg" deltaphi="360" lunit="mm" name="complicate_volume_approx" rmax="1000" rmin="0" startphi="0" z="2000"/>
</solids>

<structure>
  <volume name="complicate_volume">
    <materialref ref="G4_Pb"/>
    <solidref ref="complicate_volume_approx"/>
    <auxiliary auxtype="cadmesh" auxvalue="geometry/complicate_volume.mesh"/>
  </volume>

  <volume name="logic_volume">
    <materialref ref="G4_Galactice"/>
    <solidref ref="bounding_box"/>

    <physvol name="physical_volume">
      <volumeref ref="complicated_volume"/>
      <position name="complicated_volume_position" unit="mm" x="0" y="0" z="0"/>
    </physvol>
  </volume>
</structure>
```

The aux tag loop in remollDetectorConstruction.cc can possibly replace the complicate_volume_approx G4Solid with a CADMesh solid. Not sure if that would work.

Advantages are seamless fallthrough when no CADMesh.

wdconinc:
This PR is so old that it predates the develop branch.

Anyway, at this point it can load stl files exported directly from the cad model, such as the (misplaced) vacuum vessel on the left below.

wdconinc:
Needs `delete` of solids and logical volumes in destructor.

wdconinc:
This is the output of `build/remoll macros/compare_with_cad.mac`, which downloads the necessary map files. You need CADMesh  support (see https://github.com/christopherpoole/CADMesh).

section_xz_-10_0000
section_xz_-05_0000
section_xz_00_0000
section_xz_05_0000
section_xz_10_0000
section_xz_15_0000
section_xz_20_0000
section_xz_25_0000
section_xz_30_0000
section_yz_-10_0000
section_yz_-05_0000
section_yz_00_0000
section_yz_05_0000
section_yz_10_0000
section_yz_15_0000
section_yz_20_0000
section_yz_25_0000
section_yz_30_0000

PR #84, opened 2018-05-26
Merge downstream_window into develop

wdconinc:
Some more uncommitted changes that need looking at... Doesn't seem like much is needed here.


wdconinc:
There's really nothing new in this branch that isn't already in develop through upstream_window. I'm just going to delete it.

PR #104, opened 2018-06-13
Pull Tracking peri into Tracking branch

wdconinc:
@cameronc137 Is this needed?


cameronc137:
This tracking branch was used by Tyler a few years ago to produce envelopes for our MOLLER CAD. Currently this is the only way we have to produce such envelopes - it just prints (x,y,z) information of all particles to a external .txt file which is further analyzed later depending on whether or not it ever hits the detector 28. 

This is not very portable, and it is very stale. I have asked Catherine Feldman to shift this tracking idea over to a new functionality (which would be checked into the main branch ultimately and be very non-invasive to other users who don't want to do tracking). We will likely get rid of the text file opening and closing business and shift over to either G4VTracking kinds of classes (like in https://github.com/sbujlab/dose_track), or with the Parallel World functionality that has recently been introduced.

The extra classes involved in G4VTracking stuff would likely affect runtime and be difficult to turn off right? Would it be best to try to just use the Parallel World, that is what parallel world was intended to be used as right?

This branch's functionality will be ported over in the next month, and is also tracked in the sbujlab remoll copy, so it can be deleted now, or in a month when this is done, whatever you prefer.

cameronc137:
Catherine at SBU is making some progress recreating this behavior with the QWeak style tracking, so I'll ask her to commit all of that and we can pull that (with most things opted off for default behavior). Once that is pulled in this branch can be deleted.

wdconinc:
@cameronc137 Do you have a status update on this branch? Last was that Catherine was going to put stuff in develop. Has that happened? Can this PR and branch be deleted?

cameronc137:
Those updates are now in develop (the part.tj[xyz] branches and associated tracking behavior) and are used actively in envelope production. This branch can be deleted and old geometry is not needed. This PR is no longer needed.

PR #119, opened 2018-07-17
Merge spin precession functionality into develop

wdconinc:
@cipriangal I'd like to merge this branch into develop. There are a few conflicts, but nothing that I can't resolve myself. Let me know if you have plans for this branch that may warrant keeping it separate for now.


cipriangal:
@wdconinc: I am sorry for the lack of communication. I think by now I have convinced everyone that everything is ok with the spin precession code you added so we should be good to merge.

PR #125, opened 2018-07-18
Merge dose into integration-dose branch

wdconinc:
From integration-dose this will go into develop...

(This will be a challenging one.)


cameronc137:
I plan to work on some of these conflicts. A lot of the changes I made do no need to be kept.

wdconinc:
@cameronc137 this is probably easier done by a single person, at least the first step of merging into dose.

PR #138, opened 2018-07-29
Merge bugfix optical properties into master

wdconinc:
@acfarrell @cameronc137 We should merge this branch back into master & develop so the bugfix propagates. Before that happens, any changes to be made to the branch? I don't know if there are geometry changes that may affect master (e.g. do you want to rename any of the Mainz test beam files or so)?


wdconinc:
I will delete this PR since the branch bugfix-opticalProperties was split off from develop and therefore can't be a bugfix into master. A new branch from master will be used to merge just the bugfix into master and develop.

PR #139, opened 2018-07-29
Merge bugfix optical properties into develop

wdconinc:
Dependent on #138


cameronc137:
@acfarrell: the naming of these new geometry files in develop/geometry is fine, but if we want to be consistent it would be good to make separate geometry files for the mainz benchmark test and the upcoming blackening study. These can live in a separate branch for doing these tests with whatever small modifications are needed, but that should include the develop branch as it stands now with these bugfixes incorporated, and any significant modifications to generators should be tracked somewhere to ensure reproducability.

Naming consistency would entail duplicating and separating _black and _mainz suffixes, and making sure the target and hall geometries are minimal/non-existent for simplicity.

PR #140, opened 2018-07-31
Feature envelope

cameronc137:
@wdconinc and @catherinef37 have added rudimentary trajectory storage (tracking particle positions) functionality (can be turned on and off with macro commands) - this saves the history of stepping action positions of all particles into new ev.tj[xyz] variables - it may be worthwhile to add some macro command flag to turn off this trajectory storage (just will all ev.tj[xyz] variables with the same placeholder empty value and let ROOT compress it for us)


wdconinc:
Looks fine to me. Does this mean feature-store-trajectory is not needed anymore? I think that's what it means, but correct me if wrong.

Feel free to merge the PR (if you have the permissions, I think you should) once the tests come back fine.

cameronc137:
Oh, I had forgotten about that branch.
It looks like these are not the same, and there is at least one thing that needs to merge into this branch first - the flag in TrackingAction to only store trajectories for the primary electrons, which is what we want to do most of the time anyway, except for some few cases where we want to track ~MeV photons, so this may be something worth setting some macro command flags for and having a default setting to store trajectories for nothing.

cameronc137:
Never mind - the two branches are similar, but feature-store-trajectory is printing the trajectories from the remollGenericDetector.cc which is now moved to remollEvent.cc and stored in part branch variables as intended, and the cut on parent particles only is in place still. 

This means that doing an envelope for photons generated from the target will need to rely on a new photon generator that treats the target cell as a lightbulb and tracks primary particles travelling in straight lines - simple enough to achieve even with electrons on target if we turn off the magnetic fields.

This passed all checks and works, so I will merge into develop and delete the old feature-store-trajectory branch.

This branch will continue to live on, as we will use it to develop and track the envelope fitting code if we do want to include it as a feature in remoll.

PR #144, opened 2018-08-01
Should we even merge feature-optical into develop?

wdconinc:
@cameronc137 Is this even needed anymore since bugfix-opticalProperties seems to have a bunch of the same? If not, then I'd suggest deleting the branch feature-optical entirely. What say you, sir?


wdconinc:
Issues to resolve are:
- remollGenBeam has SetOriginXYZ where develop got SetPositionXYZ from feature-opticalProperties
- different geometry in mollerMother_black.gdml... you should assess.

cameronc137:
Great, thanks! I would assume that this is all obsolete now, but that is up to @acfarrell to determine

acfarrell:
I've been switching around the geometries in mollerMother_black.gdml while running some checks, so I don't have a preference of which one is there for now. I plan on reorganizing the geometry files in a more coherent way, so the specified geometry will probably be changed soon anyway. 

In regard to the beam generator, I think setting the hit positions is more intuitive than setting the origin positions, but this version hasn't been tested very much yet so it's not perfect.

PR #146, opened 2018-08-02
Pull old sbujlab/dose_track branch into JeffersonLab/cameron_dose_branch so geometry updates are tracked

cameronc137:
I think I corrected the large file moller_field_maps problem.

(documented git steps in the commit message for the previously bad commit, also, pushing a modified but similar version of a branch to overwrite contents can be done with "git push -u originName +HEAD")


wdconinc:
No more large file anymore. Thanks. I approve.

PR #147, opened 2018-08-02
Merge Final roof for downstream collimator area into develop 

wdconinc:
Old dangling branch by @devika123 @rakithab. Should this be merged? If not, I'd delete the branch.


wdconinc:
Conflict is likely just a CRLF mismatch. Someone should run dos2unix on Latech branch.

cameronc137:
This looks like it has been since passed by with the _merged geometry in the geometry folder, but it looks fine to keep to me.

wdconinc:
Conflicts are clearer now. @cameronc137 take another look

cameronc137:
@rakithab Is it ok to delete this branch, or would you like to keep it around so it can be backtracked to a point before the remoll v2.0.0 data structures were implemented? This would probably be needed to maintain usability of old analysis scripts that referred to this branch's data structures before the develop-merge, without updating them. If you would like to see a working variant of your old rad_dose code see www.github.com/sbujlab/rad_analysis (rad_dose_v2.cc and Makefile that builds it).

PR #149, opened 2018-08-15
Fixing trajectory storage to work for all possible generators' parent particles

cameronc137:
Trajectory storage was looking for trackID < 2 (was intended to be <= 2, for moller generator purposes) and was then storing all trajectoryContainer[i]->getPoint().x/y/z information into a part branch leaf. We have changed it to check whether the trajectoryContainer[i]'s initial momentum magnitude is == part.p variable (within 0.001 MeV) and then assign a new part.trid variable to this. There is no change in functionality, but now one can define some hypothetical generator with an arbitrary number of mother particles, all of which will get their own unique trajectories successfully stored.

Potential RAM or hard disk storage changes:
- New part.trid variable
- Properly segment part.tj[xyz] information into the correct part branch positions so no data is duplicated
- A few new conditional checks that could take a bit of time frequently

### DIFF
--- a/include/remolltypes.hh
+++ b/include/remolltypes.hh
@@ -49,6 +49,7 @@ struct remollEventParticle_t {
   double sx, sy, sz;
   double th, ph, p;
   double tpx, tpy, tpz;
+  int trid;
   std::vector<double> tjx, tjy, tjz; //Trajectory information
 };
 
--- a/macros/runexample_envelope.mac
+++ b/macros/runexample_envelope.mac
@@ -3,7 +3,7 @@
 /remoll/setgeofile geometry/mollerMother_merged.gdml
 # This must be explicitly called
 /run/initialize
-/remoll/tracking/set 3
+/remoll/tracking/set 0
 /tracking/storeTrajectory 1
 /remoll/addfield map_directory/blockyHybrid_rm_3.0.txt
 /remoll/addfield map_directory/blockyUpstream_rm_1.1.txt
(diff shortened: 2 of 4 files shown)


wdconinc:
I don't like the arbitrary 0.001 MeV difference check. Is it needed? Why isn't `GetParentID() == 0` enough?

wdconinc:
The only other request would be to fill part.trid for all tracks, not just some of them. But yeah, I see the issue there, though.

wdconinc:
I guess I was going to suggest relying on the fact that `idx` in that `for (size_t idx = 0; idx < fPartType.size(); idx++)` loop at https://github.com/JeffersonLab/remoll/blob/1adf20e7d82d06a7d3729bd52eb54605ee8acd83/src/remollEvent.cc#L42 was always going to be the primary track ID. Then I think you can just set `part.trid` to `idx` outside the conditional and use `(*trajectoryContainer)[i]->GetTrackID() == idx`.

Once the correct trajectory is found we can probably also `break` out of the `for(G4int i = 0; i < trajectoryContainer->entries(); i++)` loop.

wdconinc:
But this all starts to be a bit ugly due to the fact that remollEvent is not a G4Event... Really wish that was possible.

cameronc137:
Yes, that works, and now the part.trid is initialized to just be idx+1 instead of 0, so we can fill the trid variable for all parts that exist (only ever done for primaries anyway, in the generators) easily.

This assumes that the order of the tracks in the trajectoryContainer will always be in exactly the same order as the trackIDs being set in the event/stepping actions (or wherever it happens in Geant), so as long as that doesn't change later or have race conditions then we are fine. Does this survive multi-threading?

wdconinc:
Yeah, I think the order will always be maintained, and yes, this survives multi-threading because a single event runs in a single thread.

PR #151, opened 2018-08-22
Merge feature geom boundary into develop

wdconinc:
@cameronc137 Ready to merge when you think this is ready. Seems fine to me.


wdconinc:
@cameronc137 This branch can be deleted, right?

Issue #158, opened 2018-11-15
Optical photons no longer reflect as expected


cameronc137:
### Environment: (where does this bug occur, have you tried other environments)
- branch (master for latest released): develop here and develop in https://github.com/sbujlab/remoll/ fork
- revision (HEAD for most recent): HEAD
- OS or system: CENTOS 6.9
- Special ROOT or Geant4 versions? 5.34 and geant4.10.04.p02
### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. clone a clean copy of the remoll (either version JeffersonLab or sbujlab), go to develop branch, cmake and make
2. execute ./build/remoll macros/blackening_external.mac
3. open the output file in root and check the number of optical photons hitting the PMT cathode with T->Draw("hit.det","hit.det==50701 && hit.pid==0") and the number of electrons hitting the quartz tile with T->Draw("hit.z","hit.pid==11 & hit.det==50001")

2. Alternatively execute ./build/remoll and within remoll run /control/execute macros/vis.mac
3. Shoot several particles at the quartz and see that in fact no photons internally reflect
(In the sbujlab version of the repo the beam generator movement macro commands are different, so this may take some finessing, or just shift the detector_5open.gdml sub-file to the +x direction by 975 mm and shoot the standard beam generator at it)

### Expected Result: (what do you expect when you execute the steps above)
When I do the same thing in the sbujlab version of the repository (a stale old branch that has been manually fastforwarded some) I see ~ 24 PEs on the photocathode per electron hitting in the quartz.

### Actual Result: (what do you get when you execute the steps above)
The photons do not internally reflect in the quartz.
I see 0.1 PEs per electron on quartz, and those are probably coming from other sources.

cameronc137:
The issue lives somewhere in here: https://github.com/sbujlab/remoll/commit/b4d61daac96efc1724a5515e5fbb7ea1147496b2

cameronc137:
The bug lives specifically in between commits b8fec16 and 8bc00f7

The 14 commits in between those two do not successfully compile for me, so I cannot say which commit introduces the exact issue.

I have made two new branches, one that still works, and the other that does not, and I include in those branches everything required for exemplifying this problem (in visualizer).

[...]

PR #163, opened 2018-11-19
Fixing optical physics by demoting parallel world physics list to the…

cameronc137:
I think this fixes issue #158 but it may break parallel world physics.

Commenting out line 63 in src/remollDetectorConstruction.cc is most likely irrelevant and probably should get put back in...


cameronc137:
@rakithab can you check to see if this change has any affect on the neutron hits bug that you were seeing the other day?

cameronc137:
@wdconinc - Unfortunately my approach to fixing this issue by making parallel world physics optional does not fix the problem that optical physics breaks when parallel physics is enabled. This means that the pion_optical_response.mac macro is broken and cannot work (though it was already broken before I tried fixing things anyway).

What should we do? Allow parallel physics to break optical, or see if there is some more detailed solution and try to make them work simultaneously?

Also, Rakitha found his bug with the neutrons, and it was due to some other gdml issue (though I still don't quite understand why only he was seeing the problem).

wdconinc:
Who do you say that pion_optical_response.mac is broken and cannot work? I have a student looking at data generated from that macro that has both optical photons and hits on detector 28...

wdconinc:
I am fine with getting rid of parallel world, but that means no more detector 28 or other virtual planes. The parallel world was a way to keep those around while we implement detailed material geometry.

cameronc137:
Yes, optical photons still work, but the reflectivity properties were messed up. It would be worthwhile to check that your simulation is treating the photons the same with and without the parallel world physics list enabled.

cameronc137:
Do the physics lists need to be put after /run/initialize or before?

wdconinc:
@cameronc137 can this branch bugfix-opticalReflections-fix be deleted? It is fully merged and we are tracking the optical<>parallel issues in #158 and develop branch.

PR #164, opened 2018-11-20
Feature cmake analysis in subproject

cameronc137:
This allows for compilation of sub-projects in the analysis folder. 

First pass cmake .. and make will just compile primary remoll, and subsequent cmake .. and make will build all subdirectories mentioned in /analysis/CMakeLists.txt


analysis/pe/CMakeLists.txt
```diff
@@ -23,14 +23,14 @@ ROOT_GENERATE_DICTIONARY(
 
 #----------------------------------------------------------------------------
 #
-include_directories(${PROJECT_SOURCE_DIR}/include)
+include_directories(${PROJECT_SOURCE_DIR}/include ${REMOLL_INCLUDE_DIR})
```
cameronc137: This addition is probably not necessary

analysis/pe/CMakeLists.txt
```diff
@@ -44,7 +44,8 @@ target_link_libraries(pe-lib ${ROOT_LIBRARIES} ${REMOLL_LIBRARIES})
 
 add_executable(pe-bin pe.cc)
 set_target_properties(pe-bin PROPERTIES OUTPUT_NAME pe)
-target_link_libraries(pe-bin pe-lib)
+target_link_libraries(pe-bin ${REMOLL_BINARY_DIR}/libremoll.so pe-lib)
```
cameronc137: This addition is needed for the executable to have access to remoll's data classes

cmake/modules/Findremoll.cmake
```diff
@@ -0,0 +1,14 @@
+# - Find remoll library
+# This module checks if remoll was compiled
+# It defines:
+# remoll_FOUND               If the remoll is found
+
+find_program(remoll NAMES remoll
+             PATHS ${PROJECT_SOURCE_DIR}/build/remoll)
```
cameronc137: Making this a module is probably overkill

wdconinc:
Yeah, that's probably because you are on ROOT 5.34. As you know, I don't have a lot of sympathy or that :-)

wdconinc:
But let me see what I can do later...

wdconinc:
I wish I could be more helpful with supporting ROOT 5.34 but I can't even get it to build with even the oldest compiler on my system every time I try...

cameronc137:
Ok. I am able to compile ROOT 6.10 now, so it isn't a deal breaker for me, but I'm not sure who else relies on older ROOT versions. This is something that should be discussed at the next simulation meeting probably.

PR #173, opened 2018-12-23
Adding in user optical physics list

cameronc137:
This duplicates the already implemented "optical" list, but now the constituent processes are laid bare, allowing for the user to edit which optical processes are included. This is critical for performing blackening studies in the light guides and to determine the relative impact of scintillation vs. cherenkov.

I also ran some simple tests turning on and off cherenkov and scintillation in the user optical code, and compared also against the "optical" list to verify satisfactorily that these two lists are giving consistent results and that both scintillation and cherenkov have been included this whole time in the "optical" list.

This may still have the issue where parallel world can't be loaded at the same time (probably because these lists overwrite eachother, so that may require condensing all user lists into one class that allows multiple being turned on and off simultaneously - this UserOptical class will be a good groundwork to start from).

### DIFF
--- a/src/remollUserOptical.cc
+++ b/src/remollUserOptical.cc
@@ -1,0 +1,116 @@
+#include "G4LossTableManager.hh"
+#include "G4EmSaturation.hh"
+#include "remollUserOptical.hh"
+
+remollUserOptical::remollUserOptical(G4bool toggle)
+    : G4VPhysicsConstructor("Optical")
+{
+
+  theWLSProcess                = NULL;
+  theScintProcess              = NULL;
+  theCerenkovProcess           = NULL;
+  theBoundaryProcess           = NULL;
+  theAbsorptionProcess         = NULL;
+  theRayleighScattering        = NULL;
+  theMieHGScatteringProcess    = NULL;
+
+  AbsorptionOn                 = toggle;
+}
+
+remollUserOptical::~remollUserOptical() { }
+
+#include "G4OpticalPhoton.hh"
+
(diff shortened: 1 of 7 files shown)


cameronc137:
This resolves #173

cameronc137:
This will put an additional constraint on users ability to pick their own G4 version. See the discussion in https://github.com/JeffersonLab/qsim/issues/2

A detail with the GetParticleIterator() method requires G4 version > 4.10.02, which may be more constrictive than we want at this time.

src/remollUserOptical.cc
```diff
@@ -0,0 +1,116 @@
...
+
+    pManager = particle->GetProcessManager();
+    if (!pManager) {
+       std::ostringstream o;
+       o << "Particle " << particleName << "without a Process Manager";
+       G4Exception("remollUserOptical::ConstructProcess()","",
+                    FatalException,o.str().c_str());
+    }
+
+    if(theCerenkovProcess->IsApplicable(*particle)){
```
cameronc137: These lines here are the ones to comment out if you want to turn off Cherenkov or Scintillation

wdconinc:
Of course, in addition to `/process/optical/processActivation Cerenkov false` you can also use `/process/optical/processActivation Scintillation false`, or the same with OpWLS, OpAbsorption, OpBoundary, OpRayleigh, OpMieHG. I'd recommend running with one enabled only, and seeing if that achieves what you wish.

wdconinc:
Proposal: close without merge, delete branch due to functionality in stock optical.

PR #176, opened 2019-01-02
Split out a fraction of cmake lines into modules

wdconinc:
In particular:
- BuildType gets set up by default as RelWithDebInfo
  but Debug if a .git directory is present (i.e. developers)
- CompilerFlags are handled consistently between build types
- Prefix is also set to defaults

Some rearranging in the CMakeLists.txt because the Geant4 use file
overrides the cflags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 #----------------------------------------------------------------------------
 # Setup the project
 #
-cmake_minimum_required(VERSION 2.8.11 FATAL_ERROR)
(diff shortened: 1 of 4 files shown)


wdconinc:
Will leave this here for a bit to allow for testing on some other platforms. This now requires cmake 3.1 or higher (essentially already required by ROOT v6 requirement).

wdconinc:
Superseded by 6aa12a39e37aec8c61d7058130072867a9151c1b which achieves what this branch does.

Issue #172, opened 2018-12-21
Kryptonite material energy deposit is wrong


rakithab:
### Environment: (where does this bug occur, have you tried other environments)
- develop branch
-  revision HEAD
-  OS or system:
-  Special ROOT or Geant4 versions? geant4.10.03 and root-6.08.00

### Steps to reproduce: (give a step by step account of how to trigger the bug)
Add a volume with Kryptonite material

### Expected Result: (what do you expect when you execute the steps above)
the energy deposit in the volume should match the energy of incident particles as Kryptonite absorb all the energy into it.

### Actual Result: (what do you get when you execute the steps above)
I do not see 100% energy deposited in the volume. see attached plots
![moller_det28_hit_e](https://user-images.githubusercontent.com/609625/50354673-e5d47600-0511-11e9-8f17-e00e6413f6f7.png)
![moller_det28_hit_edep](https://user-images.githubusercontent.com/609625/50354674-e5d47600-0511-11e9-803c-a3564c879904.png)

cameronc137:
I think that the hit.edep variable is still computed as a one-of-many steps in a volume kind of parameter, so it may be that it is simply calculating the energy deposited in that first step, and since after that step is done the entire track is killed, it doesn't continue stepping and depositing the rest of its energy as it would in a non-kryptonite scenario. Is this behavior here new, or has it always been this way? 

It may be possible to modify the way the kryptonite functionality is implemented to set the momentum to zero during that initial step, and then kill the track afterward, so that the edep calcuator sees that the energy has been absorbed by the material.

wdconinc:
So I guess the problem here is the following:
- when using full materials (e.g. G4_Pb), which is slow, we get the following hits in a block of materials:
```

**Comment by wdconinc** [comment]
@rakithab Can you confirm my analysis above?

**Comment by wdconinc** [comment]
One might also wonder why in the simulation with kryptonite disabled the sum of hit deposited energies does not agree with the sum object...
```
Lead
[...]

PR #182, opened 2019-01-15
Bugfix to issue #172. Now kryptonite works again.

wdconinc:
Since this is a quite major set of changes to how some of the hits
are being stored, I am putting this in a bugfix branch first so I
can test optical photon hit behavior and let @cameronc137 take a look.

In any case, here's hit behavior:
- by default only hits are stored that satisfy the following:
  - charge particle, energy > 0.1 MeV
  - anywhere along a primary track
  - first hit in volume for secondary track
- all hits (even those not stored) are added to the detector sums

To modify this behavior you can use
```
/remoll/SD/type 4001 lowenergyneutral
/remoll/SD/type 4001 secondaries
```
along with similar GDML statements (also possible: `opticalphotons`
and `boundaryhits`).

As for kryptonite, this was all moved to remollDetectorConstruction
and stepping action was removed (since now not useful). Here is the
code for kryptonite:
```
/remoll/kryptonite/verbose 1
/remoll/kryptonite/add G4_Pb
/remoll/kryptonite/list
/remoll/kryptonite/enable
/remoll/kryptonite/disable
```
(This is different from previous `/remoll/kryptonite/set true` since
we need a method and those don't work with `G4GenericMessenger::DefineMethod`
since a G4bool looks like a G4int and yada yada yada.)

Finally, kryptonite now acts by setting all volumes that have the
specified material in question with userlimits to (0,0,0,inf,inf).
This means that all energy is deposited in the first hit, making
the sums correct. Essentially it sets the userlimits to:
```
/remoll/geometry/userlimits/usermaxallowedstep krypt_logical 0.0*mm
/remoll/geometry/userlimits/usermaxtracklength krypt_logical 0.0*mm
/remoll/geometry/userlimits/usermaxtime krypt_logical 0.0*ns
/remoll/geometry/userlimits/userminekine krypt_logical 1000*GeV
/remoll/geometry/userlimits/userminrange krypt_logical 1000*km
```
for those volumes. It does mean that steplimiter must be enabled
by default. And if you disable steplimiter and still try to use
userlimits or kryptonite there is no warning... so don't do that.

### DIFF
--- a/analysis/issues/issue172.C
+++ b/analysis/issues/issue172.C
@@ -25,6 +25,12 @@ void Analyze(TChain* T)
 
 void issue172()
 {
+  TChain* lead_all = new TChain("T");
+  lead_all->Add("issue172_lead_all.root");
+
+  std::cout << "Lead (all)" << std::endl;
+  Analyze(lead_all);
+
   TChain* lead = new TChain("T");
   lead->Add("issue172_lead.root");
 
--- a/include/remollDetectorConstruction.hh
+++ b/include/remollDetectorConstruction.hh
@@ -6,6 +6,7 @@
 #include "G4Types.hh"
 
 #include <vector>
+#include <set>
 
 class G4Tubs;
 class G4LogicalVolume;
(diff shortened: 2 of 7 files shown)


wdconinc:
Bugfix to issue #172

PR #200, opened 2019-02-19
Merge branch feature-no-default-kryptonite

wdconinc:
This removes all default kryptonite materials from the hard-coded executable and replaces them with a similar setup in macros/kryptonite.mac to be included (using /control/execute). Existing scripts that use `/remoll/kryptonite/set true` will fail (intentionally, since this is substantially different behavior).


wdconinc:
Something is wrong here... `build/remoll macros/tests/unit/test_physlist_steplimiter.mac` suddenly had 11k tracks / event (times out travis) where develop has only 50 tracks / event.

wdconinc:
Ah, I see, kryptonite is enabled by default (I did not know that)... so this of course causes lots of showering in that particular test case.

This means that any macro without `/control/execute macros/kryptonite.mac` will see a (substantial) increase in the running time (and potentially secondary showering in sensitive detectors). Again, this will only be introduced in the next release cycle leading up to v2.2, not in the v2.1 release I'm preparing now.

PR #201, opened 2019-02-20
Pull develop into a 2 week old stale branch, and then merge these few updates together

cameronc137:
Some updates to envelope code so other people can do it too


cameronc137:
Closing this PR, because the branch works as is and there are a bunch of things we need to do with it before merging with upstream modifications to detectors and kryptonite.

PR #204, opened 2019-02-25
Merge geometry pruning branch into develop

wdconinc:


wdconinc:
@cameronc137 Can this be merged?

cameronc137:
Yes, this is fine. However I think it would be good for us to establish a standard procedure for annotating any given simulation results with the commit hash used to generate them (especially if the gitinfo containing root files are not kept around after the fact) and establish a standard procedure for replication studies, both with the old geometry and old source code, and as verification of consistent source code by pulling the old geometry and running with HEAD source code. This should probably be made a separate issue and probably should feature as either a discussion or a declaration at the next MOLLER Collaboration meeting. Clearly the gitinfo is already implemented, so most of this would just be a question of teaching users how to best take advantage of that and establishing standard reproducibility studies techniques.

wdconinc:
Agreed. Maybe we need a simple command like `AddTagToCanvas(TCanvas*)` inside `reroot` or so to put a grayed out branch:sha[7] on the plot or so.

PR #205, opened 2019-02-25
Merged all matrices and materials files into a single of each type.

wdconinc:
Rationale: there's no difference between the materials. If they are
not used in a particular geometry, they can still be defined. Optical
properties can be defined without turning on optical physics.

Still includes the NIST materials data base copies which probably
can go away at some future point.


wdconinc:
PR fails due to subtargetregion in develop not having matrices.xml. Will fix after merge.

PR #206, opened 2019-02-25
Geometry include detector positions in positions.xml

wdconinc:
Move the position block outside the copies of mollerMother.gdml and include as ENTITY. This means that a repositioning is only a matter of changing a single file.


wdconinc:
@chandabindu @cameronc137 The conflict in mollerMother_merged.gdml is:
```
<<<<<<< geometry-include-detector-positions
    <physvol>
    <file name="targetDaughter_Clamshell_Optimized.gdml"/>
    <positionref ref="targetCenter_Clamshell"/>
    <!--<file name="targetDaughter_merged.gdml"/>
    <positionref ref="targetCenter"/>-->
    </physvol>
=======
      <physvol>
	      <!--<file name="targetDaughter_Clamshell_Optimized.gdml"/>-->
      <file name="subTargetRegion.gdml"/>
      <positionref ref="targetCenter_Clamshell"/>
      <!--<file name="targetDaughter_merged.gdml"/>
      <positionref ref="targetCenter"/>-->
      </physvol>
>>>>>>> develop
```

The targetDaughter_Clamshell_Optimized geometry was in develop when I branched a few days ago, but because of @chandabindu 's geometry commits in the bugfix-save-all-hits branch, there is now subTargetRegion in develop. What should we have in mollerMother_merged? I don't care too much since I think mollerMother.gdml is what should be the default, but still.

Also, this means that targetDaughter_Clamshell_Optimized changed into subTargetRegion with a pull request earlier today.

wdconinc:
Tracking remaining issue in #210

chandabindu:
Hi Wouter,
The targetDaughter_Clamshell_Optimized.gdml has the old target design (from
E158 experiment). We will not use this for MOLLER. Rakitha did all his
optimization.
The subTargetRegion.gdml contains the proposed MOLLER target geometry and
shielding. This file we will use in future.

Issue #199, opened 2019-02-19
Travis doesn't actually fail on failed unit tests...


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- branch (master for latest released): develop
- revision (HEAD for most recent): HEAD
- OS or system: JLabCE docker container
- Special ROOT or Geant4 versions? JLab CE 2.3

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. Trigger a build on `develop`, wait until it fails to complete macros/tests/unit/test_kryptonite.mac

### Expected Result: (what do you expect when you execute the steps above)
Since there is a non-existent command in that unit test, it should fail and report the build as failing.

### Actual Result: (what do you get when you execute the steps above)
It happily continues to chug along, https://travis-ci.org/JeffersonLab/remoll/jobs/495578718#L8026

wdconinc:
Fixed in 661485526047e847be89704863169a25de37a92c, hopefully.

My local system tends to deadlock on a segfault in messengers... That's also a failure mode in travis but one that is caught already by timing out :-)

wdconinc:
Not solved. remoll exits with exit code 0 even if a macro contains incorrect lines.

wdconinc:
https://travis-ci.org/github/JeffersonLab/remoll/builds/685474453 should have been marked as failed. It may be that `time` is causing the failing script to show as success. It may be that `set -euo pipefail` is not sufficient in the script.

PR #207, opened 2019-02-25
Bugfix issue 199

wdconinc:
This should fix the issue that travis doesn't indicate some test macro failures, as reported in #199. Submitting in a separate branch to trigger a travis build and test case to inspect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 script:
 - docker build -t jeffersonlab/remoll .
 - for suite in unit commit release ; do
-    docker run -t jeffersonlab/remoll "scripts/tests/test_macros.sh ${suite}" ;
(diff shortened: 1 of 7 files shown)


wdconinc:
Apparently this branch bugfix-issue-199 was not branched off develop... Oh well. Changed the base to that branch.

wdconinc:
Sure did fail on the first run through now :-) #199 fixed.

cameronc137:
What do you mean that libremoll.so isn't used? Is this a travis specific problem?

wdconinc:
Where did I write that? libremoll.so is used. This was just a bug in the test script that caused it not to actually fail when the specific tests failed.

cameronc137:
Sorry, I misread IO as so 👍

PR #212, opened 2019-02-28
Release candidate v2.1

wdconinc:


wdconinc:
This is the release candidate branch for v2.1. It is a snapshot of develop as of 2/28/2019, 10am. Continued development can go on in develop. Release candidate bug fixes only should be applied in the release branch directly or (preferably, in case of larger bug fixes) as a bugfix branch off of this release branch.

wdconinc:
Checks required:
- [ ] figure out why there's a conflict here
- [ ] look at all the plots generated by macros/tests/commit, compare with previous release
- [ ] look at all the plots generated macros/tests/release, compare with previous release

wdconinc:
The following are the standard plots for this release candidate and for the previous release (current master):
[commit.v2.1.0-alpha6.analysis.pdf](https://github.com/JeffersonLab/remoll/files/2984884/commit.v2.1.0-alpha6.analysis.pdf)
[commit.v2.0.8.analysis.pdf](https://github.com/JeffersonLab/remoll/files/2984886/commit.v2.0.8.analysis.pdf)

I'm working on getting these to actually have meaningful titles, but ok, not holding up a release for that... Essentially the files are what you get with:
```
./scripts/tests/test_macros.sh commit
cd rootfiles/tests/commit/analysis
pdftk *.pdf cat output commit.v2.1.0-alpha6.analysis.pdf
```
I'm also hoping to get pdftk into the docker container so we can just produce these pdf collections as a matter of course.

Notable differences are:
- Beam generator has a raster that generates a beam at 7.75m upstream of the target, where the raster has a focus. Almost all extent of the raster is in the angle there, and therefore not visible in vx and vy plots.
- Due to the showermax there is a lot more background in the virtual detector plane of det 28. This is now split into two plots, one for non-gammas and one for gammas, and both the forward and backward moving particles are plotted (unfortunately not in a different color).
- A set of power deposition histograms are produced by the beam generators, so that adds a set of new plots that can't be compared.

Other notes:
- None of these plots are rate weighted (but probably should be).

wdconinc:
@cameronc137 or @cipriangal I am ready for merging this into master when one of you approves it. One that's done I can continue work in develop and this can be pushed out as a standard version with which all results at the collaboration meeting were obtained.

cameronc137:
Sorry, I haven't had a chance to deploy this exact version, but the version of develop I was using from yesterday was fine (for optical simulations to benchmark SLAC beam test data), so if that counts then I give it my approval. 

I can also run a few test cases later today and verify the other things are working from this specific candidate if you'd prefer.

wdconinc:
There shouldn't be many differences between develop and the release candidate.

wdconinc:
I was holding off on merging this until I had a chance to evaluate the reported slowdowns in #252. That has been identified as not related to physics and not causing wrong results (still annoying, I agree, and still should be fixed as a bugfix release against v2.1). So, I'll go ahead and release this as agreed.

PR #217, opened 2019-03-01
update range to accept -180 deg. give more information on fail

cipriangal:
small change in how we process the header for the magnetic field. The configuration before did not allow the map to start at -180 degrees.

### DIFF
--- a/src/remollMagneticField.cc
+++ b/src/remollMagneticField.cc
@@ -219,16 +219,29 @@ void remollMagneticField::ReadFieldMap(){
     // Sanity check on header data
 
     if( !( fMin[kR] >= 0.0 && fMin[kR] < fMax[kR] &&
-	 	-180.0 < fMin[kPhi] && fMin[kPhi] < 180.0 && 
-	 	-180.0 < fMax[kPhi] && fMax[kPhi] < 180.0 && 
-	 	fMin[kPhi]  < fMax[kPhi] &&
-		fMin[kZ] < fMax[kZ] &&
-		fN[kR] > 0 && fN[kPhi] > 0 && fN[kZ] > 0 &&
-		fNxtant > 0
+	   -180.0 <= fMin[kPhi] && fMin[kPhi] < 180.0 && 
+	   -180.0 <= fMax[kPhi] && fMax[kPhi] < 180.0 && 
+	   fMin[kPhi]  < fMax[kPhi] &&
+	   fMin[kZ] < fMax[kZ] &&
+	   fN[kR] > 0 && fN[kPhi] > 0 && fN[kZ] > 0 &&
+	   fNxtant > 0
        	 )
       ){
 	G4cerr << "Error " << __FILE__ << " line " << __LINE__ 
 	    << ": File " << fFilename << " contains invalid header data.  Aborting" << G4endl;
+	G4cerr<<"fMin[kR] >= 0.0: "<<bool(fMin[kR] >= 0.0)<<" "<<fMin[kR]<<G4endl
+	      <<"fMin[kR] <= fMax[kR]: "<<bool( fMin[kR] <= fMax[kR])<<" "<<fMax[kR]<<G4endl
+	      <<"-180.0 < fMin[kPhi]: "<<bool( -180.0 < fMin[kPhi])<<" "<<fMin[kPhi]<<G4endl
+	      <<"fMin[kPhi] < 180.0: "<<bool(fMin[kPhi] < 180.0)<<G4endl
+	      <<"-180.0 <= fMax[kPhi]: "<<bool(-180.0 <= fMax[kPhi])<<" "<<fMax[kPhi]<<G4endl
+	      <<"fMax[kPhi] < 180.0: "<<bool(fMax[kPhi] < 180.0 )<<G4endl
+	      <<"fMin[kPhi]  < fMax[kPhi]: "<<bool(fMin[kPhi]  < fMax[kPhi] )<<G4endl
+	      <<"fMin[kZ] < fMax[kZ]: "<<bool(fMin[kZ] < fMax[kZ] )<<" "<<fMin[kZ]<<" "<<fMax[kZ]<<G4endl
+	      <<"fN[kR] > 0: "<<bool(fN[kR] > 0 )<<" "<<fN[kR]<<G4endl
+	      <<"fN[kPhi] > 0: "<<bool(fN[kPhi] > 0 )<<" "<<fN[kPhi]<<G4endl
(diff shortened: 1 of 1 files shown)


wdconinc:
Can we make the test inclusive of +180 deg too?

cipriangal:
If I understand correctly how G4/remoll process fields, my only concern with that is that we could have -180 **and** 180 and then we have a "infinitesimally" small region of field doubling!?

wdconinc:
My worry is that if we specify a field map from -180 to, say, +179 we will have a small region without field... :-)

I will have to look in more detail.

cipriangal:
For the regions where we don't have a field (say between -180 and 179) the code interpolates afaik. We also use the 7 fold symmetry most of the time with these field maps.

wdconinc:
I think this interpolation works exactly because of the n-fold symmetry (n gt 1). First the point is rotated to the 360/n extent around zero, then looked up in the map grid, which is by design larger than 360/n (it always prints a warning even). For n = 1 it will not do the rotation and doesn't know about the wrap around. That's what I remember. Been a while I looked in that code.

rahmans1:
Just wanted to mention that I did assume that it generalises to n=1. I haven’t looked too deeply into the code either.

wdconinc:
There's indeed a gap between -180 and +177 degrees. Here are values at the phi = 90*k +/- epsilon with the "symmetric" fieldmap:
```
At r(0.00037,0.014,9.55) [m]: B = -6.25988e-12 1.03429e-13 -8.50884e-16 T
At r(0.00037,-0.014,9.55) [m]: B = 6.25988e-12 1.03433e-13 8.50634e-16 T
At r(0.014,0.00037,9.55) [m]: B = -8.5734e-14 5.98926e-12 -6.10655e-13 T
At r(-0.014,0.00037,9.55) [m]: B = -1.85002e-13 -6.23779e-12 8.36787e-17 T
At r(-0.00037,0.014,9.55) [m]: B = -6.22348e-12 -2.13503e-13 -2.27239e-16 T
At r(-0.00037,-0.014,9.55) [m]: B = 6.22348e-12 -2.13502e-13 2.27465e-16 T
At r(0.014,-0.00037,9.55) [m]: B = 0 0 0 T
At r(-0.014,-0.00037,9.55) [m]: B = 1.85007e-13 -6.2378e-12 -8.35279e-17 T
```

For comparison here's the phi = 90*k equivalent in the default -25 to +25 deg maps:
```
At r(0,0.16,9.55) [m]: B = -5.04231e-11 4.75071e-12 4.27685e-12 T
At r(0,-0.16,9.55) [m]: B = 5.04231e-11 4.75072e-12 -4.27685e-12 T
At r(0.16,0,9.55) [m]: B = 0 0 0 T
At r(-0.16,0,9.55) [m]: B = 3.19205e-18 -5.64926e-11 -2.30044e-18 T
```
but we know that we are missing a part there...

You can get this output with
```
build/remoll macros/tests/unit/test_fields.mac
```
after removing the comment characters there.

@cipriangal This means that field maps have to either specify the +180 degrees points as well (they are identical to the -180 degrees points) or we have to make the code smarter so it knows it's a periodic coordinate.

wdconinc:
@cipriangal I started fixing this in the code, but it essentially becomes a fix for only this particular field map. I think it would be better to regenerate the field map to not have a gap between -180 and +180 degrees.

wdconinc:
```
#!/bin/bash

head -n 6 $1
{
  tail -n +7 $1
  tail -n +7 $1 | awk -v OFS="\t" '{ if ($2==-180) { $2=180; print($0) } }'
} | sort -n -k3,3 -k2,2 -k1,1
```

cipriangal:
@rahmans1 : I think @wdconinc is right here. We should probably just set a particular definition for how we export files from TOSCA and have only one implementation. Could you get the offset and symmetric maps to not have a gat between -180 and 180?

wdconinc:
I can upload the field maps with the additional lines added to the moller web location. I'll just call them 0.1 instead of 0.0 (assuming that was supposed to be a version number...).

wdconinc:
(I'm referring to those listed here: https://hallaweb.jlab.org/12GeV/Moller/downloads/remoll/)

wdconinc:
Updated map with that snippet gives me:
```
G4WT0 > At r(0.00037,0.014,9.55) [m]: B = -6.25988e-12 1.03429e-13 -8.50884e-16 T
G4WT0 > At r(0.00037,-0.014,9.55) [m]: B = 6.25988e-12 1.03433e-13 8.50634e-16 T
G4WT0 > At r(0.014,0.00037,9.55) [m]: B = -8.5734e-14 5.98926e-12 -6.10655e-13 T
G4WT0 > At r(-0.014,0.00037,9.55) [m]: B = -1.85002e-13 -6.23779e-12 8.36787e-17 T
G4WT0 > At r(-0.00037,0.014,9.55) [m]: B = -6.22348e-12 -2.13503e-13 -2.27239e-16 T
G4WT0 > At r(-0.00037,-0.014,9.55) [m]: B = 6.22348e-12 -2.13502e-13 2.27465e-16 T
G4WT0 > At r(0.014,-0.00037,9.55) [m]: B = 8.5729e-14 5.98926e-12 6.10655e-13 T
G4WT0 > At r(-0.014,-0.00037,9.55) [m]: B = 1.85007e-13 -6.2378e-12 -8.35279e-17 T
```
so it does not have the gap.

rahmans1:
@wdconinc The reason why I avoided generating between -180 to 180 was I thought the magnetic field is added cumulatively in overlapping regions. However, it would only be the case for that one point I guess... so negligible effect? Simple fix would be to just add an extra line for theta =180 field value that has the same value as theta=-180 for each z-coordinate which you have already done it seems....

wdconinc:
The field will not double-count itself even if it wraps around several times in azimuth (but if you load multiple fields defined in the same region of space they will add up). You could theoretically define a field map with a phi grid between 0 and 720 degrees (maybe for a spin 1/2 field?) and it would still return the value (in this case the appropriate value between 0 and 360 degrees).

I have indeed added the lines for +180 equal to those for -180, resorted the file, and fixed the header to reflect this. I don't have other maps, but if you do you should be able to fix them with that little bit of code too.

wdconinc:
Uploaded to https://hallaweb.jlab.org/12GeV/Moller/downloads/remoll/. Will update CMakeLists.txt to refer to those (with correct MD5). Will update remollMagneticField to complain about these kind of gaps in the field.

The problem is primarily that this is right in the middle of an open sector :-)

rahmans1:
@wdconinc Thanks. Sounds reasonable to me.

PR #223, opened 2019-03-15
Added targz_macros.sh script to pack up the test products (pngs, root, log)

wdconinc:
then copy these out of the container and deploy as release products.


wdconinc:
[ ] Needs merging into develop too!

Issue #231, opened 2019-03-19
Bug: remollGenPion fails assert in SamplePhysics


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- branch (master for latest released): release-candidate-v2.1
- revision (HEAD for most recent): HEAD
- OS or system: travis
- Special ROOT or Geant4 versions? JlabCE:2.3

### Steps to reproduce: (give a step by step account of how to trigger the bug)
Not sure if this is easily reproduced, but wanted to record it for posterity.

### Expected Result: (what do you expect when you execute the steps above)
Generator should always pass assertions in the code.

### Actual Result: (what do you get when you execute the steps above)
Log at https://api.travis-ci.org/v3/job/508475535/log.txt
```
remoll: /jlab/remoll/src/remollGenPion.cc:68: virtual void remollGenPion::SamplePhysics(remollVertex*, remollEvent*): Assertion `pf < beamE' failed.

/jlab/remoll/scripts/tests/test_macros.sh: line 63:   846 Aborted                 (core dumped) $prefix build/remoll ${macro} 2>&1
```

wdconinc:
Same in https://api.travis-ci.org/v3/job/508476825/log.txt

wdconinc:
Maybe related to #120

wdconinc:
Not sure why this actually popped up suddenly. Of course, once it pops up it remains an issue since we start from the same random seed...

wdconinc:
I see... Travis (and JLabCE) runs in single threaded mode. https://api.travis-ci.org/v3/job/508524633/log.txt... Still fails tests then.

PR #233, opened 2019-03-19
Run pion test simulation with single thread

wdconinc:
This works around issue #231 (but doesn't solve it).


wdconinc:
Not able to reproduce locally even with fresh Docker build on JlabCE.

Issue #242, opened 2019-03-22
Change default raster reference z to center of target, change correlation to reflect new raster location


wdconinc:
Correlations determined from 6.7 m from center of pair of raster coils to the center of the target.

wdconinc:
For a +/- 2.5 mm raster, that's a correlation of 149 urad / mm (so 2.5 mm translates into 373 urad angle). That's up from 136 urad / mm for a raster pair 7.75 m in front of the target, indeed realistic.

wdconinc:
Closing, further discussion in PR #243.

PR #243, opened 2019-03-22
Update default beam gen raster origin

wdconinc:
Update default beam gen raster origin to -6.7 m, corr to 149 urad/mm, refz to 0.0 m

Fix to issue #242.

Raster is now by default 5 mm x 5 mm full size at the center of the
target (z = 0.0 mm) with an angle/position correlation of 149 urad/mm
corresponding with origin at -6.7 m per Jay's new designs.

### DIFF
--- a/src/remollGenBeam.cc
+++ b/src/remollGenBeam.cc
@@ -21,16 +21,16 @@
 
 remollGenBeam::remollGenBeam()
 : remollVEventGen("beam"),
-  fOriginMean(0.0*m,0.0*m,-7.75*m),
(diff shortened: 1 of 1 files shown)


rakithab:
Raster is now by default 5 mm x 5 mm full size at the center of the
target. Is that true? In previous experiments raster full size was the US of the target, I'm just checking.

wdconinc:
At the March 2019 collaboration meeting, Jay indicated that the raster calculations all assume a raster at the center of the target. Previous experiments used a much shorter target so it wasn't as important.

This PR just sets the default raster Z position to the center of the target. You can still change it with a macro command.

PR #244, opened 2019-03-22
Changed target length to 1.25 m

wdconinc:
This propagates the constants
```
    <constant name="tubeTarget_length" value="1250"/>
    <constant name="AlWindowTarget_length" value="0.127"/>
```
and placement of the windows
```
      <physvol>
        <volumeref ref="USAlTarg"/>
        <position name="targ_upstream_window" unit="mm" x="0" y="0" z="-(AlWindowTarget_length+tubeTarget_length)/2"/>
      </physvol>

      <physvol>
        <volumeref ref="DSAlTarg"/>
        <position name="targ_downstream_window" unit="mm" x="0" y="0" z="+(AlWindowTarget_length+tubeTarget_length)/2"/>
      </physvol>

```
in all target implementations.

Asides:
- the `/remoll/targlen` command is likely to disappear since changing
this currently won't move the aluminum windows.

### DIFF
--- a/geometry/target/subTargetRegion.gdml
+++ b/geometry/target/subTargetRegion.gdml
@@ -10,6 +10,9 @@
   <define>
     &matrices;
 
+   <constant name="tubeTarget_length" value="1250"/>
+   <constant name="AlWindowTarget_length" value="0.127"/>
+
     <constant name="targetChamber_R" value="1981.0/2.0"/>
     <constant name="targetChamber_L" value="1829."/>
     <constant name="targetChamber_thickness" value="13."/>
@@ -64,10 +67,10 @@
    </subtraction>
 
    <tube aunit="deg" lunit="mm" deltaphi="360" name="target_solid"
-	  rmax="40" rmin="0" z="1500"/>
+	  rmax="40" rmin="0" z="tubeTarget_length"/>
 
     <tube aunit="deg" lunit="mm" deltaphi="360" name="AlWindow"
-	  rmax="40" rmin="0" z="0.127"/>
+	  rmax="40" rmin="0" z="AlWindowTarget_length"/>
 
     <tube aunit="deg" lunit="mm" deltaphi="360" name="targetChamber_solid1"
 	  rmin="targetChamber_R"
@@ -266,7 +269,7 @@
 
      <physvol>
        <volumeref ref="USAlTarg"/>
-       <position name="targ_upstream_window" unit="mm" x="0" y="0" z="-750-1"/>
+       <position name="targ_upstream_window" unit="mm" x="0" y="0" z="-(AlWindowTarget_length+tubeTarget_length)/2"/>
      </physvol>
 
      <physvol>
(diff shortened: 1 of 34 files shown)


wdconinc:
As another aside, the aux tag `<auxiliary auxtype="TargetSamplingVolume" auxvalue=""/>` makes this target volume eligible to be a sampling volume (you still have to select it in the macro). In `geometry/target/subTargetRegion.gdml` these are currently disabled for aluminum windows so US and DS Al target simulations will not work with these.

wdconinc:
Still need to delete /remoll/targlen and /remoll/targpos.

cipriangal:
This looks good to me.

wdconinc:
Move entire target center downstream by +0.5 m. This means raster should also be referenced to +0.5 m.

wdconinc:
Origin vertex now as per image below.

image

cipriangal:
@wdconinc : if we have +(AlWindowTarget_length+tubeTarget_length)/2 won't these two volumes share a surface? I remember that used to be a problem. Otherwise it looks fine to me.

wdconinc:
I have never run into problems with those, as long as they are well defined boundaries on primary coordinates. Is there a geant4 recommendation to avoid these kind of shared surfaces?

Exceptions for optical processes (since boundaries are special there) and world volume which can't have shared boundaries with physical volumes.

cameronc137:
I feel like we should try to keep the origin defined at the center of the target, since otherwise now it has a seemingly arbitrary 500mm offset and the raster is being moved anyway. It would require changing all of the UOFFSET variables -500 and I think that would be it? It probably doesn't matter, but 5 years from now it may be unclear why a half meter shows up everywhere.

cipriangal:
@wdconinc : maybe I am mis-remembering (couldn't find anything too troubling from a quick search online). 
@cameronc137 : it would cause more problems right now to move the subTargetRegion by half a meter. We will need to revisit this anyway in the summer and at that point we can move the whole thing. If you have the time to fix all the overlaps and problems that will come up I'd be happy to move the whole Region.

cameronc137:
I mean to keep the entire target still wrt the origin, then move the raster (if needed, based on Jay) and move every other sub volume by 500mm in the other direction. And yes this would probably be a mess, so we don't need to worry about it now.

jmammei:
Should be the pivot.



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

cipriangal:
we should discuss it at the next Sim mtg. I tend to agree with @jmammei, in particular as we will be moving to constructing things and placing them inside the Hall CAD model we will run into translation issues when discussing with engineers. 

For now we should just move the target volume as @wdconinc  has already done and we can change the paradigm later.

wdconinc:
I think standard pivot at zero is the way to go (agree with Juliette). Helps in interactions with hall tech staff. Also agree with Cip to do this later, not now: we only want to make this change once.

Issue #221, opened 2019-03-14
Add large acceptance fields to cmake download list


wdconinc:
Locations:
https://hallaweb.jlab.org/12GeV/Moller/downloads/remoll/hybridSymmetric_sensR_0.0.txt
https://hallaweb.jlab.org/12GeV/Moller/downloads/remoll/upstreamSymmetric_sensR_0.0.txt
MD5 sums:
0f6e9a3a8392d01c63f1609afc4e4c09  hybridSymmetric_sensR_0.0.txt                                                                                                                   
4ca3d968cadd85a89b86148411eda678  upstreamSymmetric_sensR_0.0.txt                                                                                                                 

**Is your feature request related to a problem? Please describe.**
Current field maps are limited in r.

**Describe the solution you'd like**
For simulations of low energy tracks at larger r, we need fields at larger r.

**Describe alternatives you've considered**
We could simply direct users to the website or add a make target to download additional fields. However, downloading directly is not as much overhead for 18Mb fields.

wdconinc:
@cipriangal I'll let you decide which maps are appropriate to upload and add here.

rahmans1:
@cipriangal Could you please let me know upto what radius in r you want the field maps to be generated in? Also the level of coarseness in azimuthal direction? We can increase the number of steps in theta for the outer fringe or leave it the same as in the main region.

wdconinc:
Resolved and working with #338.

PR #245, opened 2019-03-22
Add cmake option ADDITIONAL_FIELDS to download more fields

wdconinc:
This addresses issue #221 by adding the following
```
cmake -DADDITIONAL_FIELDS=ON ..
```
which will download upstreamSymmetric_sensR_0.0.txt and
hybridSymmetric_sensR_0.0.txt into map_directory.

Tried to solve this by introducing a separate make target but this was
faster and ultimately equivalent in documentation requirements.


wdconinc:
Once the functionality is there we can easily add more publicly accessible fields with their md5 checksum.

PR #249, opened 2019-04-02
Bugfix: Set default build type to RelWithDebInfo

wdconinc:
The default build type was Debug, which is about ~30% slower than
RelWithDebInfo, and RelWithDebInfo contains the same backtrace info.
This will make code faster (since typically run without changing the
default build type).

A specific build type can be chosen without the neeed to specify a
command line flag by creating a file .buildtype (which is ignored by
git) with as content e.g. Debug. This will persist between deletes
of the build directory (but changing the file will not cause a new
build type to take effect, this is only read on the first call to
cmake in an empty directory). The content of the file can still be
overridden with a command line argument: e.g.
```
cmake -DCMAKE_BUILD_TYPE=Release ..
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,9 @@ log
 .cproject
 .project
 
+#ignore build type file
+.buildtype
+
 #ignore dot files
 *.dot
 
--- a/cmake/modules/BuildType.cmake
+++ b/cmake/modules/BuildType.cmake
@@ -1,7 +1,10 @@
 # Set a default build type if none was specified
 set(DEFAULT_CMAKE_BUILD_TYPE "Release")
 if (EXISTS "${CMAKE_SOURCE_DIR}/.git")
-  set(DEFAULT_CMAKE_BUILD_TYPE "Debug")
+  set(DEFAULT_CMAKE_BUILD_TYPE "RelWithDebInfo")
+endif()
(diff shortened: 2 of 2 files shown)


wdconinc:
Check how much improvement in speed between RelWithDebInfo and Release?

wdconinc:
@cipriangal Here's the impact on runexample.mac with 10k events (similar load on system, but probably +/- 10 s real time based on past experience):

Debug:
```
real    7m18.218s
user    14m21.795s
sys     0m0.928s
```

RelWithDebInfo (default):
```
real    5m17.848s
user    10m21.468s
sys     0m0.833s
```

Release:
```
real    5m16.444s
user    10m19.447s
sys     0m0.817s
```

The remaining difference between RelWithDebInfo and Release is indeed in the file size (there is also MinSizeRel, which optimizes for smaller binaries even at the expense of execution speed):
```
-rwxrwxr-x 1 wdconinc wdconinc 1.3M Apr 11 15:44 build_release/libremoll.so
-rwxrwxr-x 1 wdconinc wdconinc  65K Apr 11 15:44 build_release/remoll
-rwxrwxr-x 1 wdconinc wdconinc  20M Apr 11 16:17 build_relwithdebinfo/libremoll.so
-rwxrwxr-x 1 wdconinc wdconinc 595K Apr 11 16:17 build_relwithdebinfo/remoll
```

Issue #241, opened 2019-03-22
Change default beam current to 70 uA


wdconinc:
We should change the default beam current and all beam current settings in macros to 70 uA. We probably should warn users when trying to run with currents higher than that.

wdconinc:
Related issue #222

wdconinc:
Relevant line: https://github.com/JeffersonLab/remoll/blob/43925db5c72678b6d70092f9163c9521201c47cc/include/remollglobs.hh#L18

wdconinc:
Approved at Simulation meeting March 28, 2019

wdconinc:
Further discussion (if any) in PR #251.

PR #251, opened 2019-04-02
Changed default beam current to 65 uA (was 75 uA)

wdconinc:
If you don't specify a beam current, it will determine rates based on 70 uA assumption.

You can still override with `/remoll/beamcurr 85 uA` etc. No changes beyond a single character...


wdconinc:
This resolves #241.

wdconinc:
Change to 65 uA by default.

Issue #10, opened 2015-09-03
Implement new raster in all branches


seamusriordan:
The new raster with an origin upstream needs to be implemented in all branches.  In particular, making sure the beam generator includes those effects is critical.

PR #262, opened 2019-05-05
Enhancement pivot is new origin

wdconinc:
Moving all geometry (rather trivially) to a new geometry origin. Since geometry is tree-based and top-level absolute positions are in `geometry/positions.xml`, this is a fairly easy.

- Some absolute positions are also in the beam generator raster (and in remollBeamTarget raster if one uses the angle correlations there, not touched here since the raster origin at -8000 m is somewhat arbitrary and needs a rewrite as part of #10).

I think that's all, though.

### DIFF
--- a/geometry/mollerMother_5open.gdml
+++ b/geometry/mollerMother_5open.gdml
@@ -26,7 +26,7 @@
 
     <physvol>
       <file name="detector_5open.gdml"/>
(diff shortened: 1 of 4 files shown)


wdconinc:
Won't merge. Will merge #263 directly.

PR #263, opened 2019-05-07
Enhancement updated geometry

wdconinc:
This merges in the updated geometry in https://github.com/JeffersonLab/remoll/projects/12.

1. target goes 1.5 m downstream and is shortened to 1.25 m
2. upstream and hybrid both moved 1 m downstream
3. detectors move 0.5 m upstream

Distance between center of target and detector plane is 26.5 m.

Raster size referenced to center of target. Effective raster magnet position downstream by 1m

There is a remaining (large) volume overlap in mollerMother_merged.gdml. See below:
image (1)
The rest seems ok.

### DIFF
--- a/geometry/mollerMother_5open.gdml
+++ b/geometry/mollerMother_5open.gdml
@@ -26,7 +26,7 @@
 
     <physvol>
       <file name="detector_5open.gdml"/>
-      <position name="detectorCenter" x="0" y="0" z="0.0*28500."/>
+      <position name="detectorCenter" x="0" y="0" z="0.0"/>
       <rotation name="rot" unit="deg" x="0" y="90" z="0"/>
     </physvol>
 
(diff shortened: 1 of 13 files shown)


wdconinc:
This PR will require that #262 is merged first since it is branched off that.

wdconinc:
The remaining overlap is now in all in irrelevant regions:
image
Compare this image with the one above to see the difference.

@cameronc137 @rahmans1 @cipriangal and whoever else: take note of 1806cc8,
> Avoid overlap with 75% less poly at front end, remove padding  …
> The main goal here is to get this merged in. The concrete shielding of
> the target region is overlapping with the upstream spectrometer poly
> shielding by about 75 cm. This commit fixes the overlap in an adhoc way.
> 
> Rather than removing concrete DS shielding on the target, I've chosen to
> remove US shielding on the upstream spectrometer in the hope that this
> will have less impact. A proper study is obviously needed.

cameronc137:
This looks old to me. Eventually we will want to tune the shielding US if col2

PR #267, opened 2019-08-29
Preparing for v2.2: merge develop into rc1

wdconinc:
This is in preparation for remoll v2.2 which will include the updated geometry (finally), and remove default kryptonite.


wdconinc:
Didn't expect so many conflicts... I must have done something wrong.

PR #275, opened 2019-09-18
Bugfix: remove content of user stepping action

wdconinc:
Removed the content of stepping initialization (was not use, was old) ￼…

This just referred to the old way of doing kryptonite and wasn't compiled in anyway. Better to have the files reflect what is actually compiled and included in the running.


wdconinc:
This should have been filed against develop, not master.

PR #276, opened 2019-10-02
Bugfix: Fix event generator to be in the target, not downstream (MAJOR)

wdconinc:
Due to a sign error, the events were generated around +4587 mm instead
of around -4587 mm. The call to GetFrameTranslation should have been a
call to GetTranslation, or various equivalent sign changes.

This was not a problem before since the target was centered at zero, and
so -0 is the same...

### DIFF
--- a/src/remollBeamTarget.cc
+++ b/src/remollBeamTarget.cc
@@ -125,7 +125,7 @@ void remollBeamTarget::UpdateInfo()
     if (!fTargetMother) {
       return;
     }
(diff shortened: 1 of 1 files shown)


rahmans1:
I haven't noticed this issue. Which branches were affected by this and was this error introduced recently or has it always been the case?

wdconinc:
This doesn't affect the beam generator so you may not have been affected. The bug here was introduced when the geometry was changed in May.

I do think everyone was probably affected by the absence of the magnetic field shift.

PR #284, opened 2019-10-07
Feature: Save full random state to ROOT file

wdconinc:
Sometimes you want to inspect (or be able to repeat) specific events that you find in a ROOT file. This branch enables (optional) saving of the full random engine state for each event.

By telling geant4 to store the random state for each event with `/run/storeRndmStatToEvent 1` it will automatically be stored to the tree as the variable `seed`. With reroot or (ld pre-)loading libremoll.so you can then use things like:
```
T->Draw("seed.Save()","ev.A>-10")
```
which will save the random state for every event that has an asymmetry with magnitude less than 10 ppb to a file run[runno]evt[evtno].rndm. Arbitrary selection possible.

The files can be turned into input for /random/resetEngineFrom <filename>. See below for comments on file format.

### Seed != state
The seed determines the random state uniquely but is in a significantly smaller space than the random state itself. Not every random state can be associate with a seed.

### File formats
The ROOT tree Save() command produce the following file 
```
MixMaxRng-begin 97588365
1906709568702062078
1140743473972090589
1823734812527682176
692017379249797100
1447898696895639102
636884972592620537
2152014336500572859
1570012997546763742
1959162295989761165
84311882159462073
355167662739815000
1006438779200467577
834618078282138689
2207380892026575602
1613299806778549612
205787593014985317
1060740182943464238
1
2250179337412895848
MixMaxRng-end
```
but the input for /random/resetEngineFrom requires the following format:
```
mixmax state, file version 1.0
N=17; V[N]={1906709568702062078, 1140743473972090589, 1823734812527682176, 692017379249797100, 1447898696895639102, 636884972592620537, 2152014336500572859, 1570012997546763742, 1959162295989761165, 84311882159462073, 355167662739815000, 1006438779200467577, 834618078282138689, 2207380892026575602, 1613299806778549612, 205787593014985317, 1060740182943464238}; counter=1; sumtot=2250179337412895848;
```
Some massaging is necessary...

You can get the correspondence by comparing the output of running with
```
/random/setSavingFlag 1
/random/saveEachEventFlag 1
```
(which will create random state file for every event, clutter your directory, and slow down disk access) with the format stored with the Save() command.

### Impact on running and file size
Below the benchmarks on runexample.mac with 1000 events with and without saving random state. First with saving state to the output file:
```
/run/storeRndmStatToEvent 1
running time 51.7981 s
output size 10802014 b
```
versus default option of not storing to file:
```
/run/storeRndmStatToEvent 0
running time 52.4119 s
output size 10588667 b
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,13 @@
 images
 
+# random seed files
+G4Master_run*.rndm
+G4Worker?_run*.rndm
+G4Worker?_run*evt*.rndm
+run*evt*.rndm
+currentEvent.rndm
+currentRun.rndm
+
 # output from createHTML
 _remoll_*.html
 
--- a/include/remollEventAction.hh
+++ b/include/remollEventAction.hh
@@ -35,9 +35,6 @@ class remollEventAction : public G4UserEventAction
       fPrimaryGeneratorAction = action;
     }
 
-  private:
-    // Random seed at begin of event
-    G4String fEventSeed;
 };
 
 #endif
--- a/include/remollIO.hh
+++ b/include/remollIO.hh
@@ -48,7 +48,7 @@ class remollSeed_t: public TObject {
     TString fSeed; //< random engine state, a.k.a. seed (but not really)
   public:
     // Default constructor
-    remollSeed_t(): TObject() { };
+    remollSeed_t(): TObject() { fRunNo = 0; fEvtNo = 0; fSeed = ""; };
     // Copy constructor (not implemented)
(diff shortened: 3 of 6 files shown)


wdconinc:
Unfortunately the output and input formats for the random states are not guaranteed to be constant over time (and not everyone is guaranteed to run with the same engines), so I am not going to provide a conversion script since it would be too fragile. On the other hand, turning 17 rows with 1 number into 1 row with 17 numbers is an easy enough exercise.

PR #266, opened 2019-07-07
Feature envelope improve

cameronc137:
I've made some changes to the analysis codes, and made a new version of reader that is done manually.

PR #286, opened 2019-10-10
Shortened updated collimators

cameronc137:


cameronc137:
I mistakenly closed this yesterday. I believe that this set of commits correctly updates the collimator definitions with the envelope-defined sculpts in the new-shortened and translated coordinate system geometry. This is important for any new simulation efforts to use this updated collimator sculpt.

cameronc137:
This does not include the lintel definitions, as those are still being worked on by Sakib.

cameronc137:
Are the travis failures from this branch serious problems?

wdconinc:
No, the travis error are due to the containers (short lived bug that was fixed in develop after this branch was branched). Just check that it does the rest correctly :-)

cameronc137:
OK. I'm reluctant to merge this directly into develop as just some commits, as it actually represents a significant change to the geometry (i.e. this set of collimator changes will change the flux reaching the main detectors from the prior 28 meter long geometry to the new 26.5 meter one used in recent tests by myself and Sakib). 
What is the mechanism for doing a new release number? Is a number like 3.0.0 appropriate?

wdconinc:
If we want an entirely new version then I think we should first merge all other minor feature branches (if any), tag a last v2, make a plan for which other geometry changes may be major enough, try to fold this into a new major version.

cameronc137:
Can we follow up on this? I think it was ready to merge, and it may have been superceded by whatever work Sakib has done recently?

wdconinc:
Sorry, no description in this PR. What was this about again?

cameronc137:
I had a few updates on top of PR #266 that need to get into the main branch for our collimator definitions to be correct in all parallel copies of the geometry. Again this shows the problem of having multiple copies of the "same" geometry with only small changes between them without good version control.

I had made changes to the collimators in PR #266 that I hope have been built upon for subsequent studies, but that only affected the set of files I had directly worked off of. This brings (probably now stale) the other copies of those files up to the same place.

Probably this should get lumped into the discussion of updating the geometry that Sakib made some commits on from Thursday #334

wdconinc:
Status?

rahmans1:
@wdconinc The collimator definitions at least in the merged suffixed files in the current develop are correct and carefully tracked. All future changes should be done on top of that. I am not sure we still have the necessity to propagate 5 parallel versions of the same geometry instead of just maintaining the correct one. We should discuss this at the sim meeting and this can probably be closed. Anyways, the conflicts here are from more recent work superceding it. This branch seems rather old.

wdconinc:
@rahmans1 Status update or can this be closed? Going through some old remoll todo lists...

rahmans1:
> @rahmans1 Status update or can this be closed? Going through some old remoll todo lists...

This issue can be closed I think. New work by Chandan in a different branch supercedes this.

PR #289, opened 2019-10-22
Feature: geometry with full parametrized detector

wdconinc:
This pulls in the remoll-detector-generator repository for the full parametrized detector model. After git pull, you may need to go into geometry/detector and git pull, depending on your setup. You will also need to run the generator before this will work.

Finally, all this is still positioned wrong.


wdconinc:
image

wdconinc:
If #296 is merged into develop first, then there some additional fixing to do here.

PR #291, opened 2019-11-12
Feature add sb sbunker

wdconinc:


wdconinc:
@chandabindu @cipriangal Does this need to be merged back into develop? If so, what is left to complete in this branch before doing so? If no, is there anything useful here that we should somehow cherry-pick out of it?

cipriangal:
I need to spend some time to see what is in here. I am fairly sure this is where all the geometry that we used for the Apr review lives so we should not delete it. Give me till the end of the week.

cipriangal:
This is the geometry we used for the target and Coll 1,2 shielding calculations. It should be preserved. The top hat configuration which I think is in develop is not accurate. I'll try to work on the conflicts this afternoon and push a commit for review.

cipriangal:
merged develop. took develop changes for positions and mollerParallel. took most of changes in target and US from this branch.

cipriangal:
@wdconinc : I give up ... currently this branch is consistent with develop from a couple of days ago. Someone removed the shielding in the US region and it's an empty region with some detectors. I will not remove our shielding/detector configurations from this branch by merging in develop. Until I understand why the shielding is removed I would like this branch to remain as is so that we can work on radiation studies if we need to.

wdconinc:
Status?

wdconinc:
As discussed in the meeting today (@cipriangal) this needs to get merged in. There are conflicts. Proposed plan of action:
1. What's needed in this branch? It also includes new lintels, changes to target shielding, and changes to hybrid region. If we only want SBS, then someone should extract only SBS into another branch.
1. We should merge current develop into branch and reasses.

Other questions:
1. Why the change in src/remollDetectorConstruction.cc?
1. Why another new geometry file moller_clamOptHallDet.gdml? Can this be in a standard file?

Is the right way to proceed maybe just to cherry-pick the geometry/electronics bits?

wdconinc:
Relevant SBS commits affecting geometry/electronics/subSBSbunker.gdml are (in order):
- a67d5a04ee69475fb5b46b9082290838b7c5c4f5
- fec95abb5f304dc5e9419d644d7fd03c3879fad1
- fefb4a6528defccc3d95f95af4268d46c12d4bff

Then, the following doesn't actually change geometry/electronics/subSBSbunker.gdml, just indents:
- ae9ee71254d3934488230ce20e889b7e142d5e92
but it does change the positioning in geometry/moller_mergedHallDet.gdml (which we can live without, I guess):
```
       <physvol>
     	<file name="electronics/subSBSbunker.gdml"/>
-	<position name="sbsBunker_pos" x="8740" y="-3000+1980" z="6087.-15180."/>
+	<position name="sbsBunker_pos" x="8740" y="-3000 + 2000" z="6087. - 15180."/>
     	<rotation name="rot_y30" unit="deg" y="30"/>
       </physvol>
```

My suggestion in the short term is:
1. cherry-pick the first three commits into a new branch that starts at da542e93f4f1ae03738da727c196577a07b6dddc, the branch point where feature-add-SBSbunker starts
2. add electronics bunker to mollerMother_merged.gdml
3. merge branch into feature-add-SBSbunker
4. merge branch into develop

wdconinc:
What's possibly left after #354 is:
- new analysis scripts and simulation macros: propose cherry-pick into new branch and merge
- mother volumes should explicitly set visibility attributes: propose cherry-pick into new branch and merge
- new parallel world virtual detector planes at alternative positions: unclear
- changes to subTargetRegion.gdml: unclear

I think that's it, but needs more detailed inspection after partial merges into develop.

wdconinc:
@cipriangal Are we ready to toss this branch out?

PR #293, opened 2019-11-13
Feature: /remoll/kryptonite/volume <logical_volume_name>

wdconinc:
This sets the logical volume to kryptonite by name.

This avoids the need to do all the kryptonite by material type.


wdconinc:
Ready for merging at next meeting.

PR #296, opened 2019-11-27
Feature: Remove all schema copies, replace local references with standard url

wdconinc:
We used a local copy of the GDML schema since it used to not be accessible on
the ifarm. That has since been changed, and it is now accessible. Instead of
copying all these files along, it is just better to replace the local copy with
a link to the remote location. We never had any modifications anyway (none that
aren't implemented in the current 3.1.6 schema anyway).

The other thing we removed here is the xmlns:gdml namespace. Not needed since
we link to the schema directly.


wdconinc:
If #289 is merged into develop first, then there some additional fixing to do here.

PR #299, opened 2019-12-03
Changed hybrid 2 bounce shield radii

rahmans1:


chandabindu:
The dimensions for twobounce shield are okay.

Issue #322, opened 2020-03-25
BeamTarget FIXME needs investigation (?!)


cipriangal:
This is an old FIXME left over from Seamus's time. Apparently sometimes (at least with the G4 version he was using then) the material->GetAtomsVector() returned a NULL. His work around is to assume a weight of 1 for that case. As we are using more and more materials I assume this could be important.

Once this is investigated we should remove the FIXME (through a cursory check of the issues I never saw this addressed). 

### Environment: (where does this bug occur, have you tried other environments)
- branch (master for latest released): develop
https://github.com/JeffersonLab/remoll/blob/develop/src/remollBeamTarget.cc#L318

### Expected Result: (what do you expect when you execute the steps above)
GetAtomsVector not NULL

### Actual Result: (what do you get when you execute the steps above)
GetAtomsVector NULL

wdconinc:
This is essentially the difference between
```
  <material Z="1" name="LiquidHydrogen" state="liquid">
    <T unit="K" value="20.27"/>
    <D unit="g/cm3" value="0.0708"/>
    <atom unit="g/mole" value="1.00794"/>
  </material>
```
and
```
  <isotope name="H1" Z="1" N="1">
    <atom type="A" value="1.01"/>
  </isotope>
  <isotope name="H2" Z="1" N="2">
    <atom type="A" value="2.01"/>
  </isotope>
  <element Z="1" formula="H" name="Hydrogen" >
    <fraction ref="H1" n="0.9999" />
    <fraction ref="H2" n="0.0001" />
  </element>
  <material Z="1" name="LiquidHydrogen" state="liquid">
    <T unit="K" value="20.27"/>
    <D unit="g/cm3" value="0.0708"/>
    <fraction n="1" ref="Hydrogen"/>
  </material>
```

In the first case we're taking an effective approach: liquid hydrogen consists of a single atom species with effective atomic weight of 1.00794. In the second case we take an exhaustive approach (resulting in an effective atomic weight of 1.00793; close enough to the representation limit of floats).

The GetAtomsVector is null for for the first case (current default in the materials.xml definition of LiquidHydrogen used in the target volumes).

Anyway this is just an explanation of what is happening. Still not sure if assuming a weight of 1 is the right thing to do in any case, or only if we assume hydrogen...

wdconinc:
Summary: simulations were correct, but requiring properly defined materials is better so we are now going to crash out instead of assuming values for atomvec.

PR #323, opened 2020-03-26
Bugfix: use G4_lH2 instead of LiquidHydrogen

wdconinc:
Addresses #322, it was the custom definition of LiquidHydrogen that failed the FIXME. With this fix, only properly formed materials are used in the target (for hydrogen and aluminum, still using an explicit 12C isotope).

Additional benefit, this now defines the proper ionisation energy of hydrogen in the target.

### DIFF
--- a/geometry/target/subTargetRegion.gdml
+++ b/geometry/target/subTargetRegion.gdml
@@ -183,7 +183,7 @@
 <structure>
 
   <volume name="h2Targ">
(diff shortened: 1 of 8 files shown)


wdconinc:
Full list of material names at http://geant4-userdoc.web.cern.ch/geant4-userdoc/UsersGuides/ForApplicationDeveloper/html/Appendix/materialNames.html

PR #334, opened 2020-04-23
Test hevac lintelsandcollars

rahmans1:
bringing in all the changes from test-hevac-lintelsandcollars into develop. A few more things left to update, like photon collimator detector id will be pulled in with the next merge.


cameronc137:
Does this include #266 and #286 commits, or at least the same geometry?

PR #345, opened 2020-05-28
Feature: sieve implementation

wdconinc:
This implements a sieve in the upstream region. Design based on attached pdf. Distances are parameters that can be changed.

image

[MovableCollimatorAsm Drawing 1.pdf](https://github.com/JeffersonLab/remoll/files/4698297/MovableCollimatorAsm.Drawing.1.pdf)


wdconinc:
10k events of macros/tests/unit/test_sieve.mac results in different distributions at the main detector plane 28:

With sieve IN:
- main detector plane:
image
- sieve hole 'detectors' 1001, 1002, 1003:
image

With sieve OUT:
- main detector plane:
image
- sieve hole 'detectors' 1001, 1002, 1003:
image

Cutaway front view of sieve when OUT (no overlap in acceptance):
image

wdconinc:
Side view to indicate where this actually lives...
image

cameronc137:
Hi Wouter, I'm not sure whether this is useful information, but the sieve idea that I am aware of from the CAD matches what you are doing, but instead has the sieve and blocker (sieve with no holes at all) sitting slightly upstream of the collimator 1, so that they can be moved horizontally in and out of the beamline on some simple movement system (not designed yet). It can be seen in the CAD pictures upstream of col1.

Iso_ALL_clean_2

wdconinc:
Merging per discussion at optics meeting today.

PR #346, opened 2020-05-28
Feature: Macro commands to move volumes

wdconinc:
Based on the name of an already placed physical volume you can now shift and rotate that volume. Takes the effort out of editing gdml (but can introduce overlaps).

Example:
```
/remoll/geometry/absolute_position h2Targ_PV (0,0,-5)
/run/beamOn 10
/remoll/geometry/relative_position h2Targ_PV (0,0,+5)
/run/beamOn 10
/remoll/geometry/absolute_rotation h2Targ_PV (0,90,0)
/run/beamOn 10
/remoll/geometry/relative_rotation h2Targ_PV (0,30,0)
/run/beamOn 10
```
(which makes no sense, but demonstrates the point).

The OpenGL visualization needs an interaction with the display for the change to be visible.

Unnamed physical volumes (`<physvol>...`) have name of the logical volume with `_PV` at the end.

### DIFF
--- a/include/remollDetectorConstruction.hh
+++ b/include/remollDetectorConstruction.hh
@@ -110,6 +110,13 @@ class remollDetectorConstruction : public G4VUserDetectorConstruction
     void InitKryptoniteMaterials();
 
 
+  public:
+
+    void AbsolutePosition(G4String name, G4ThreeVector position);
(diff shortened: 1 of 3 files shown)


wdconinc:
The main point for this is to be able to move the sieve in and out.

PR #347, opened 2020-06-04
Feature: Add pre-moller vacuum beampipe outside the target region

wdconinc:
image


wdconinc:
WIP since still needs the matching pipe inside the target region.

wdconinc:
Depends on #348.

wdconinc:
Now looks like this:
image

wdconinc:
The beampipe in the premoller region is entirely arbitrarily set to 10cm diameter (upstream of target region) and 7cm (target region). The limit of 7cm is to avoid overlaps. If someone has values in the CAD, those can be used. @cameronc137 maybe.

wdconinc:
Verification that this works. When using beam generator at -19.81 m, and with artificially changing this back to air in the beampipe in the premoller region, there are indeed vertices upstream of the target:
image

PR #352, opened 2020-06-05
Feature: Set raster origin position to -19.81m (beam generator)

wdconinc:
This works: beam generator with vacuum premoller beampipes results in a properly sized raster in the target.
image


wdconinc:
Sorry, this is branched off develop with PR #346 merged into it...

PR #354, opened 2020-06-11
Feature: add only SBS bunker

wdconinc:
Per #291 [comment](https://github.com/JeffersonLab/remoll/pull/291#issuecomment-642852430) this merges just the SBS bunker into develop, none of the other stuff in feature-add-SBSbunker.


wdconinc:
Trivial conflict resolved.

chandabindu:
I am sure that the positions is with respect to old origin (wrt to target center). We have to change the position for the new origin (wrt hall center)..

wdconinc:
Good point. Seem z is 15180 mm upstream of hall center.

wdconinc:
image

PR #355, opened 2020-06-12
Feature: require alpha of mother volumes

wdconinc:
Until now, mother volumes (top level physical volumes) were automatically set to wireframe so we don't end up with a big white blob. This feature, by @cipriangal in feature-add-SBSbunker PR #291, requires that you put the aux tag into the gdml. Cleaner, less crappy code to manage, more power to the user.


wdconinc:
Will probably require adding aux tags to a few volumes in addition to the ones in the branch. At least branch has develop merged into it.

PR #377, opened 2020-08-31
Reduce geometry clutter

wdconinc:
Removing Mainz and 5open geometry in develop. 2 years old, not been touched. If needed @cameronc137, it'll still be here in the history.


cameronc137 (approved):
I've got another branch (feature-pe-studies) that will involve this folder, but as you say these versions are all old. I will try to make sure that all detector geometries are fully encoded in the other repository for gdml detector generation, rather than stored here in their full forms.

PR #380, opened 2020-09-12
Feature: Extract concrete/lead donut from pion system, and other geometry updates

wdconinc:
This started with just pulling the concrete/lead donut out of the pion system so we can move it independently, but when checking positions it turned out some other things were in the wrong place. That in turn resulted in pion detector movement in various radial directions too.

Summary: here is the +X view with overlaps indicated
image
- small known overlap between showermax reflectors,
- some other beam pipe overlaps which have been there for a while now.


wdconinc:
TODO
- [x] Make sure concrete/lead donut is in all mother volumes of relevance (merged and parametrized).
- [ ] ~~Remove unused mother volumes, so only merged and parametrized are left.~~
- [ ] ~~Rename mother volume merged to default without suffix.~~

Last two items moved to new branch and PR.

PR #385, opened 2020-09-19
Feature: smaller world: x="75000" y="50000" z="125000" lunit="mm"

wdconinc:
Until now the world was 200 m cubed; this reduces this to 125 m in the z direction only, 75 m left/right, and 50 m up/down.

All particles are tracked all the way to the edge of the world (which is filled with G4_AIR, not vacuum). This probably won't speed up your simulations by enormous factors, but no need to waste cycles either :-)

### DIFF
--- a/geometry/hall/hallDaughter_acceptanceDefinition.gdml
+++ b/geometry/hall/hallDaughter_acceptanceDefinition.gdml
@@ -11,8 +11,8 @@
 <define>
   &matrices;
 
-  <constant name="width_mother" value="90000."/>
-  <constant name="height_mother" value="90000."/>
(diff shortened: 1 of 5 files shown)


jmammei:
Does it still include the floor, ceiling and walls of the hall?  Isn't the hall 32 m in radius?

Or is that only needed for special radiation simulations?

Juliette



Sent from my Samsung Galaxy smartphone.



-------- Original message --------

wdconinc:
Still includes the entire hall. Doesn't include Jefferson Avenue anymore, in case someone wants to do site boundary dose simulations.

image
image

PR #387, opened 2020-09-21
Feature: strong reproducibility and random states

wdconinc:
This is a bit technical, but hopefully leads to something useful: visualizing events selected from the ROOT file.

The main issue is of course that geant4 is doing its very best to simulate events that are as random as possible. How can you then visualize a single event in a run of 10M events that you are curious about? 

Some of the issues:
- Random engine state is not stored by default in the remoll output ROOT tree: an internal geant4 command exists and is used by remoll to store the `seed` variable, but documentation could be better (or we could enable this by default at about a 1% file size hit).
- CLHEP (which is the random engine that is used) writes the random engine status in a different format to a string stream compared to a file. What is stored in the ROOT tree is therefore not immediately useful but needs to be converted. This PR introduces a converter script.

Anyway, there's now a unit test that simulates 10 events, then simulates them once more in backwards order. The unit test checks that all events indeed have the correct random engine states stored in the ROOT tree.

Next, this needs a bit of docs to aid in visualizing that one event.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ G4Worker?_run*evt*.rndm
 run*evt*.rndm
 currentEvent.rndm
 currentRun.rndm
+*.state
+*.rndm
 
 # output from createHTML
 _remoll_*.html
--- a/README.replay.md
+++ b/README.replay.md
@@ -1,0 +1,52 @@
+## Replaying simulated events
+
+These instructions explain how to replay exactly some previously simulated events.
+
+### Run the original large-N simulations
+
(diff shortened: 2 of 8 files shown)


wdconinc:
Probably needs a script that takes a collection of random  state files and queues them up for viewing. E.g.
`scripts/convert-to-run run?evt*.state` would create a bunch of sequential run?evt0.rndm files that can then be stepped through with the visual interface.

Wish there was a way to setup the simulation by setting the same generator etc.

PR #392, opened 2020-09-30
Feature: default parallel world to mollerParallel instead of blank

wdconinc:
Default to a parallel world given by geometry/mollerParallel.gdml instead of blank. This will cause regular `build/remoll` to automatically show parallel world planes as well in visualization.


wdconinc:
Now you can still use `/remoll/parallel/setfile` without argument to turn off parallel world (i.e. have an entirely empty parallel world). This is useful for those who use remoll for dedicated detector geometries for which the default parallel world is not useful.

wdconinc:
Also, this shouldn't have been so hard:
```
  fParallelMessenger->DeclareMethod(
      "setfile",
      &remollParallelConstruction::SetGDMLFile,
      "Set parallel geometry GDML file")
          .SetStates(G4State_PreInit)
          .SetDefaultValue("")
          .command->GetParameter(0)->SetOmittable(true);
```
It would seem that setting a default value would automatically set that parameter to omittable, but no. Probably should be an upstream geant4 bug report...

PR #394, opened 2020-10-01
Added remollSearchPath.hh

Yug34:
Test it [here.](https://repl.it/@Yug34/remoll#test.cc)


CMakeLists.txt
```diff
@@ -383,6 +383,14 @@ install(FILES
     ${PROJECT_SOURCE_DIR}/cmake/scripts/RemollUseFile.cmake
     DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/remoll)
 
+#----------------------------------------------------------------------------
+# remollSearchPath
+
+link_libraries(stdc++fs)
```
wdconinc: Can you find out from which c++ versions this is supported?

wdconinc:
Can you change the following classes so they uses remollSearchPath?
- [ ] remollDetectorConstruction: for `geometry/mollerMother.gdml`
- [ ] remollParallelConstruction: for `geometry/mollerParallel.gdml`
- [ ] remollGlobalField: for `map_directory/....txt`

wdconinc:
Ideally I think we should get to the point where we can do the following (starting in the toplevel remoll directory):
```
mkdir build
cmake -B build -DCMAKE_INSTALL_PREFIX=/usr/local .
make -C build
make -C build install
cd /tmp
remoll macros/runexample.mac
```
which would
- install all field maps, geometry, macros to `/usr/local/share/remoll/`
- `remoll macros/runexample.mac` would automagically find the macro in `/usr/local/share/remoll/macros/runexample.mac`

CMakeLists.txt
Yug34: I checked it for C++20, 17, 14, 11, 98. It works for all of them except C++98.

PR #395, opened 2020-10-05
Feature: pion donut frame

wdconinc:
There are questions about large radius electrons scattering off the frame. Here's a frame.

image


wdconinc:
I-beam implementation overkill? Nah.

wdconinc:
TODO Needs R = 71.356 cm hole for tapered beampipe.

wdconinc:
image

PR #401, opened 2020-10-19
[gdml] reference to defined position needs `positionref` tag

wdconinc:
This does not affect the simulation since `boxDSShield_local_trans` was anyway defined as zero. But that's only until someone decides to change it...

@rahmans1 Is this file generated from something? Does it need to be changed upstream somewhere?


rahmans1 (approved):

rahmans1:
@wdconinc This file is not autogenerated. So, it should be fine to just have the change here.

PR #405, opened 2020-11-02
[donut] updated mounting frame: rotated I-beams, now aluminum

wdconinc:
Based on discussion Nov 2, 2020, the frame looks more like this, and is made out of aluminum.

Note: these are 40 cm high I-beams, not 16.33 in as they will be in reality. In reality the I-beams will extend beyond the lead.

image


wdconinc:
I have not implemented the larger I-beam frame because it requires increasing the size of the donut mother volume. That's not as trivial and interferes with our pion detector volumes.

wdconinc:
image

wdconinc:
image

wdconinc:
Fins added
image

wdconinc:
Actually, correct sectors
image

wdconinc:
Origin vertex of secondaries when detected in pion lucite (all particle types, mostly gammas)
image

PR #411, opened 2020-11-12
updated merged and extended collimator 1-2design

chandabindu:
The updated collimator 1-2 merge design is here.

### DIFF
--- a/geometry/upstream/upstreamDaughter_merged.gdml
+++ b/geometry/upstream/upstreamDaughter_merged.gdml
@@ -10,8 +10,6 @@
 <define>
     &matrices;
     
-    <variable name="i_septant" value="0"/>
-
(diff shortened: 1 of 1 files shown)


geometry/upstream/upstreamDaughter_merged.gdml
```diff
@@ -782,40 +540,35 @@
     <volume name="logicUpstream">
       <materialref ref="VacuumColl"/>
       <solidref ref="boxUpstream"/>
+      <auxiliary auxtype="Alpha" auxvalue="0.1"/>
```
wdconinc: Please leave auxiliary tags at the end, per the GDML specification. If you put it earlier, you will get validation errors.

wdconinc (changes requested):
See comment on auxiliary tag.

PR #412, opened 2020-11-16
Feature change magnet z offset

rahmans1:
With this pull request, all prior field maps will be unusable as default since they use the old target center (5m upstream of hall center) as reference. New field maps will be provided that use the hall center as reference 0.

### DIFF
--- a/src/remollMagneticField.cc
+++ b/src/remollMagneticField.cc
@@ -42,7 +42,7 @@ remollMagneticField::remollMagneticField( G4String filename ){
 
     // Default offset for field maps in reference frame with
     // the hall pivot at z = 0.
-    fZoffset = -5000.0;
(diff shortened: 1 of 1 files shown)


wdconinc:
We can still set the offset for old field maps in a macro. We should include these instructions.

wdconinc:
In general, if an old code continues to run it should be correct. If it can't be made correct, it should crash until the user fixes the issue. We don't want users to run with old field maps and get wrong results without realizing it.

rahmans1:
> We can still set the offset for old field maps in a macro. We should include these instructions.

That's true. We will probably settle on new magnet designs very soon. However, if people want to have access to the current maps for archival purposes, i will write up the instructions on how to do that.

wdconinc:
Merged into #414

PR #413, opened 2020-11-19
Feature: new magnetic field maps

wdconinc:
This introduces the new magnetic field maps in the naming scheme per the magnetic interface definition document.

Outstanding issues:

- [ ] reduce the number of field maps downloaded by default to the minimum
- [ ] update references in the load_magnetic_fieldmap macro to the latest versions


wdconinc:
Wrong base branch.

PR #414, opened 2020-11-19
Feature new magnetic field maps

wdconinc:
This introduces the new magnetic field maps in the naming scheme per the magnetic interface definition document.

Outstanding issues:

- [x] reduce the number of field maps downloaded by default to the minimum
- [x] update references in the load_magnetic_fieldmap macro to the latest versions


wdconinc:
```
build/remoll macros/runexample.mac
root -l remollout.root
T->Draw("hit.r","(hit.det==28 && hit.pid==11)")
```
image

PR #415, opened 2020-11-26
Feature: make rate variable work with beam generator

wdconinc:
For the beam generator, the `rate` variable was filled with zero. This PR fills the `rate` variable so it works identical to the other generators.

The effective luminosity for physics process generators requires the effective target length *L &rho; N<sub>A</sub> / A* (target length L, density &rho;, Avogadro N<sub>A</sub>, atomic mass A) multiplied with the beam frequency I/e (beam current I, electron charge e) in the effective luminosity to multiply with the differential cross section &sigma; of the event times the phase space d&Omega;, all normalized over the total number of simulated events N.

When an event generator (such as the beam generator) has no target (kNoTargetVolume), we instead pass the beam frequency, I/e, with `GetEffLumin`, and pass a cross section of 1 in `fEffXs`. The product results in a weight of I/e/N per event.

### DIFF
--- a/include/remollBeamTarget.hh
+++ b/include/remollBeamTarget.hh
@@ -62,7 +62,7 @@ class remollBeamTarget {
         remollBeamTarget();
 	virtual ~remollBeamTarget();
 
-	G4double GetEffLumin();
+	G4double GetEffLumin(SamplingType_t);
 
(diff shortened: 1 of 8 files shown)


wdconinc:
Thanks to @rahmans1 for testing:
test

PR #416, opened 2020-12-04
Added BariteConcrete and Marble for shielding purposes

chandabindu:
Added new material for target region shielding purposes


wdconinc (approved):
Thanks for implementing this.

Issue #382, opened 2020-09-19
Geometry overlap internal to beampipe/downstream/beampipeDSMother.gdml


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- branch (master for latest released): develop
- revision (HEAD for most recent): HEAD 5656858ef608cc7ebfa7e934c11650011ac796b8
- OS or system: Linux ubuntu 20.04
- Special ROOT or Geant4 versions? Just the usual

### Steps to reproduce: (give a step by step account of how to trigger the bug)
1. cd geometry
2. gdmlview -o -c beampipe/downstream/beampipeDSMother.gdml

### Expected Result: (what do you expect when you execute the steps above)
Expect no overlaps.

### Actual Result: (what do you get when you execute the steps above)
```
-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume DSpipe2_DSpipe2_2:0 (G4Polycone)
          with DSpipe2_DSplate:0 (G4Polycone) volume's
          local point (531.962,-32.4457,19594.4), overlapping by at least: 99.9032 um 
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------


-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume DSpipe2_DSplate:0 (G4Polycone)
          with DSpipe2_DSpipe2_2:0 (G4Polycone) volume's
          local point (299.95,434.71,19589.8), overlapping by at least: 2.03512 um 
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------
```

wdconinc:
![image](https://user-images.githubusercontent.com/4656391/93672367-eea00600-fa6f-11ea-81fc-9062687539fe.png)

chandabindu:
Fixed!!

PR #418, opened 2020-12-08
bug-fix issue #382: beampipe/downstream/beampipeDSMother overlap

chandabindu:
Please review and merge this into develop


wdconinc (approved):
Verified that this indeed resolves the overlap. The only overlap left is in the showermax as described in #419.

PR #423, opened 2020-12-08
Feature: require cxx 17

wdconinc:
Several instances in the code already use c++17 syntax. It is aso required by recent geant4 versions. This pull request is also intended to get feedback:
- @rahmans1: Compute Canada?
- @chandabindu: Ifarm support?
- @cameronc137, @cipriangal: SBU cluster?
Just a test compilation and runexample.mac verification would be sufficient :-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,10 +20,10 @@ include(CMakeEnv)
 # Use GNU install dirs
 include(GNUInstallDirs)
 
-# Request C++11 standard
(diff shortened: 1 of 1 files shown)


chandabindu:
I just tried to build this branch in ifarm.. it gave enormous number of errors...
Here are first few lines. 
remollBuildError

wdconinc:
Thanks, @chandabindu, for the feedback. This does require some solution since at this point remoll does not compile against ROOT installed with c++17 (soon to be default).

PR #424, opened 2020-12-10
[cmake] do not force cxx standard (even though we use 17 constructs)

wdconinc:
We have to let ROOT define the cxx standard, due to the way ROOT operates (https://root.cern/install/build_from_source/#caveats).


wdconinc:
This now allows compilation against geant4 and root both compiled with c++17.

PR #433, opened 2021-01-21
[hall] det 99 closer to hall wall; hole for DS beampipe to 26.5m

wdconinc:
To avoid overlap between downstream beampipe, decided on flat interface at z = 26.5m.

This requires cutting a 770 mm radius tube out of the hall volume up to 26.5m, which is where the downstream beam pipe fits in.

It also means putting detector 99 a bit closer to the wall at radius 26517.6 mm: instead of at 26417.6 mm (which would cause issues with the 26.5 m interface), it is now at 26507.6 mm.


cipriangal:
@wdconinc : Thank you! This should be ready to be placed in develop when you think it would be appropriate.

PR #434, opened 2021-01-22
Add SAMs geometry to develop

wdconinc:
This should not have any more overlaps (after #433).

- [x] Move jlabsubmit.py out of top level


wdconinc:
@vdoomra Can we just delete jlabSubmit1.py or do you want to keep this in here? The risk is that others will use it, but not update the email address ;-) Alternative, put it in a different folder (jobs/).

Issue #454, opened 2021-03-23
remoll needs a command line option to specify filename


wdconinc:
**Is your feature request related to a problem? Please describe.**
When running large job arrays of the same macro, we still need to modify the macro to avoid having the same output file name. Being able to specify the output file name on the command line would avoid this.

**Describe the solution you'd like**
`remoll -o remollout.root`

**Describe alternatives you've considered**
We could pass a run number, or a string tag, or parse the environment for `SLURM_ARRAY_TASK_ID`. All of those are clunkier.

wdconinc:
Resolved by PR #464

PR #464, opened 2021-04-06
Added CLI option to specify output file name

Yug34:
For: [Issue 454](https://github.com/JeffersonLab/remoll/issues/454)

### DIFF
--- a/remoll.cc
+++ b/remoll.cc
@@ -101,7 +101,21 @@ int main(int argc, char** argv) {
       else if (G4String(argv[i]) == "-p") parallel_gdmlfile = argv[++i];
       else if (G4String(argv[i]) == "-u") session  = argv[++i];
       else if (G4String(argv[i]) == "-r") seed     = atol(argv[++i]);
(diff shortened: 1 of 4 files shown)


remoll.cc
```diff
@@ -98,6 +101,21 @@ int main(int argc, char** argv) {
       else if (G4String(argv[i]) == "-p") parallel_gdmlfile = argv[++i];
       else if (G4String(argv[i]) == "-u") session  = argv[++i];
       else if (G4String(argv[i]) == "-r") seed     = atol(argv[++i]);
+      else if (G4String(argv[i]) == "-o") {
+        outputfile = argv[++i];
+
+        const std::string illegalChars = "$?<>:*\?\\\'\"";
```
wdconinc: We should not strip out the characters.

wdconinc (changes requested):
We should not change runexample.mac, and the comments there are probably confusing to people who just want to use that file. 

We should just set outputfile like in line 103 above it.

PR #467, opened 2021-04-12
Feature: cubic magnetic field interpolation

wdconinc:
This introduces cubic interpolation of the magnetic fields. This should be able to reproduce fast changes in the field map more accurately. The linear interpolation was also rewritten using a convolution kernel, so that it makes more sense how the cubic interpolation works.

- At the edge of the field map (boundary layer of 1 cell size), cubic interpolation is impossible in its canonical form since it would require a point at index -1 or index `size` (where `size-1` is the largest allowable). In these cells, the cubic interpolation falls back to the linear interpolation.
- In the case of phi we can do better (this is implemented): we wrap can around from the edge at -180 to +180 (in fact, we have to, otherwise the sector at -180 will look different because it will use linear interpolation when all other sectors use cubic). This is only correct for the field maps that we are currently using: from -180 to +180. At this moment, the code will fail to produce correct results for other field maps. Either, we toss that functionality entirely, or we need to add support for sector fieldmaps to the cubic interpolation.
- In the case of r at the lower edge (r = 0, index 0) we can do better as well (this is not implemented): we can assume that the value at index -1 is the value of the field at index 1, but for 180 degrees around the azimuth. At this point, we just assume linear when close to r = 0.

For runexample.mac (100 events), we get 323720 field lookup of which 5008 are 'downgraded' from cubic to linear. When comparing directly with the linear case, there are large differences (in the ratio), but those are on very small values. It would be useful to check whether there are large differences in areas where the field is sizeable and well-behaved.

Someone can probably also run a largish simulation and compare the peak position/shape.

### DIFF
--- a/include/remollGlobalField.hh
+++ b/include/remollGlobalField.hh
@@ -27,6 +27,7 @@ class remollGlobalField : public G4MagneticField {
 
         void AddNewField(G4String& name);
 
+        void SetInterpolationType(const G4String& name, const G4String& type);
         void SetZOffset(const G4String& name, G4double offset);
         void SetFieldScale(const G4String& name, G4double scale);
         void SetMagnetCurrent(const G4String& name, G4double current);
--- a/include/remollMagneticField.hh
+++ b/include/remollMagneticField.hh
@@ -45,7 +45,8 @@ class remollMagneticField : public G4MagneticField {
 	remollMagneticField(const G4String&);
 	virtual ~remollMagneticField() { };
 
-	void GetFieldValue(const G4double Point[4], G4double *Bfield) const;
+	void AddFieldValue(const G4double point[4], G4double *field) const;
+	void GetFieldValue(const G4double point[4], G4double *field) const;
 
 	void SetFieldScale(G4double scale) { fFieldScale = scale; }
 	void SetRefCurrent(G4double current) { fRefCurrent = current; }
--- a/include/remollMagneticField.hh
+++ b/include/remollMagneticField.hh
@@ -42,24 +42,20 @@ class remollMagneticField : public G4MagneticField {
 
     public:
 
-	remollMagneticField( G4String );
-	virtual ~remollMagneticField();
+	remollMagneticField(const G4String&);
+	virtual ~remollMagneticField() { };
 
 	void GetFieldValue(const G4double Point[4], G4double *Bfield) const;
 
-	void InitializeGrid();
-	void ReadFieldMap();
+	void SetFieldScale(G4double scale) { fFieldScale = scale; }
+	void SetRefCurrent(G4double current) { fRefCurrent = current; }
+	void SetCurrent(G4double current) { SetFieldScale(current/fRefCurrent); }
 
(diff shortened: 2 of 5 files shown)


wdconinc:
Linear interpolation between two exact grid points in the 180 degree plane of the field map:
```
At r(0.35,0,8.5) [m]: B = 0.00357703 0.00050849 0.00792857 T
At r(0.3502,0,8.51) [m]: B = 0.00328225 0.00281274 0.00717497 T
At r(0.3504,0,8.52) [m]: B = 0.00297535 0.00513092 0.00642428 T
At r(0.3506,0,8.53) [m]: B = 0.00265633 0.00746304 0.0056765 T
At r(0.3508,0,8.54) [m]: B = 0.00232519 0.0098091 0.00493164 T
At r(0.351,0,8.55) [m]: B = 0.00198193 0.0121691 0.00418968 T
At r(0.3512,0,8.56) [m]: B = 0.00162656 0.014543 0.00345063 T
At r(0.3514,0,8.57) [m]: B = 0.00125906 0.0169309 0.00271449 T
At r(0.3516,0,8.58) [m]: B = 0.000879449 0.0193327 0.00198125 T
At r(0.3518,0,8.59) [m]: B = 0.000487718 0.0217484 0.00125093 T
At r(0.352,0,8.6) [m]: B = 8.38688e-05 0.0241781 0.000523521 T
```
Cubic interpolation between two exact grid points in the 180 degree plane of the field map:
```
At r(0.35,0,8.5) [m]: B = 0.00357703 0.00050849 0.00792857 T
At r(0.3502,0,8.51) [m]: B = 0.00281228 0.0032963 0.0075301 T
At r(0.3504,0,8.52) [m]: B = 0.00218551 0.0060392 0.00692629 T
At r(0.3506,0,8.53) [m]: B = 0.00168302 0.00871853 0.0061654 T
At r(0.3508,0,8.54) [m]: B = 0.0012892 0.0113158 0.00529581 T
At r(0.351,0,8.55) [m]: B = 0.000986578 0.0138129 0.00436606 T
At r(0.3512,0,8.56) [m]: B = 0.000755712 0.0161919 0.00342481 T
At r(0.3514,0,8.57) [m]: B = 0.000575104 0.0184353 0.00252086 T
At r(0.3516,0,8.58) [m]: B = 0.00042097 0.0205256 0.00170319 T
At r(0.3518,0,8.59) [m]: B = 0.000266974 0.0224456 0.00102095 T
At r(0.352,0,8.6) [m]: B = 8.38688e-05 0.0241781 0.000523521 T
```
Comparison in a plot for each of the components:
image
Comparison in a plot for the magnitude:
image

@rahmans1 A much bigger effect than I had expected...

wdconinc:
With the neighboring point as well:
image
Triangles are cubic, small points are linear.

wdconinc:
@rahmans1 Merge?

rahmans1:
Yup. Should be good to merge.
________________________________

rahmans1 (approved):

PR #471, opened 2021-04-19
[fields] New magnetic field maps per @rahmans1

wdconinc:
New magnetic field maps have been spotted on the hallaweb. Default now. Update to use.


rahmans1 (approved):
looks good

PR #477, opened 2021-04-23
Feature: target ladder implementation

wdconinc:
This implements the target ladder with its variety of targets. Targets can be selected with, e.g.:
```
/control/execute macros/target/Optics1.mac
```
This does **not** set the event generator to anything (since many generators are appropriate for various targets, e.g. the Moller generator applies to the carbon optics foils as well).

Note: The Al hole target does **not** generate a distribution with a hole when using a regular physics generator. This target is mainly intended for use with the beam generator.


wdconinc:
Reviewer @cipriangal @rahmans1 requests: please check if your usual workflow still returns valid results with this branch :-)

wdconinc:
Note: This branch currently still calculates rate incorrectly. Not sure why, but looking into it.

wdconinc:
- [ ] Remove references to /remoll/printtargetinfo in favor of new command.

wdconinc:
This branch now produces event-wise identical results for runexample.mac when compared to develop.

Waiting for a review from @cipriangal or @rahmans1 before merging.

PR #479, opened 2021-04-24
Feature: [vis] added horizontal XZ plane section view to gui

wdconinc:
image

Somehow XZ is glitchier than YZ but still very usable.


wdconinc:
Well, this is a pretty trivial removal of a single space now...

PR #480, opened 2021-04-25
Feature target shield design

cpalatchi:
update mag fields


cpalatchi:
accident - don't want to do this

PR #489, opened 2021-05-07
Feature collimator2 update

chandabindu:
Updated geometry for collimator2 to avoid edge scattering.

### DIFF
--- a/geometry/upstream/upstreamDaughter_merged.gdml
+++ b/geometry/upstream/upstreamDaughter_merged.gdml
@@ -34,8 +34,8 @@
 
     <!-- main coll, adjust this -->
     <position name="posCOLL2_CW" unit="mm" x="0" y="0" z="5925-UOFFSET-50-75-50"/> <!-- upstream end of CW piece is 5250mm with respect to target center -->
+    <position name="posCOLL2_CW_1" unit="mm" x="0" y="0" z="5925-UOFFSET-50-75-50+50"/> <!-- upstream end of CW piece is 5250mm with respect to target center -->
     <position name="posCOLL2_Cu" unit="mm" x="0" y="0" z="5925-UOFFSET-50-75+75-25"/> <!-- upstream end of Cu piece is 5350mm with respect to target center -->
-    <position name="posCOLL2_Cu1" unit="mm" x="0" y="0" z="5925-UOFFSET-50-75+75-25+50"/> <!-- upstream end of Cu piece is 5350mm with respect to target center -->
     <!--Shielding beam pipe overlap with col-2 so I increased to bore radius in col-2 from 26.0 to 29.55 -->
     <constant name="COLL2_R1_CW_U1" value ="27.5"/>
     <constant name="COLL2_R1_CW_D1" value ="27.5"/>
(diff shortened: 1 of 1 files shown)


geometry/upstream/upstreamDaughter_merged.gdml
```diff
@@ -350,88 +346,195 @@
...
+        </subtraction>
+
+        <subtraction name ="Coll2_Cu_S8">
+        <first ref="Coll2_Cu_S7"/>
+        <second ref="Coll2_Cu_S2"/>
+        <position name="pos_Cu_6" x="0" y="0" z="25" lunit="mm"/>
+        <rotation name="rot_Cu_6" x="0" y="0" z="SEPTANT*(5+0.75)" unit="deg"/>
+        </subtraction>
+
+        <subtraction name ="Coll2_Cu_S9">
```
wdconinc, line 532: The repeated subtractions are very slow, especially for collimators where many tracks are passing through. Can you implement this with nested volumes? Can this repeated subtraction live inside a vacuum volume based on the arb8 Coll2_Cu_S? At least then geant4 only has to deal with the subtractions when it is actually inside the arb8.

wdconinc: Of course, same applies to the other repeated subtraction as well. This is only possible if the arb8 has nothing 'sticking' into it.

wdconinc: @chandabindu

wdconinc (changes requested):

geometry/upstream/upstreamDaughter_merged.gdml
chandabindu: If it  was okay with the previous Collimator 2 and collimator 4 design (especially number subtraction they had in their geometries), then the present implementation of collimator 2 should also be fine - there are only one extra-subtraction step involved for the new  collimator 2 stage (though the W-part of the collimator 2  now has two solids).

I ran simulations with 10K events (with same seed) with beam generators with 'develop' and with 'feature-collimator2-update' branch on ifarm1901..
Here is the run time statistics--

**For 'develop' branch**
G4WT0 > --> Event 9000 starts with initial seeds (45799854,61822005).
### Run 0 ended (1208.71s).
Writing output to remollout.root ... done
Graphics systems deleted.
Visualization Manager deleting...
 Running time[s]: 1231.85


**For 'feature-collimator2-update' branch** 
G4WT1 > --> Event 9000 starts with initial seeds (45799854,61822005).
### Run 0 ended (1208.59s).
Writing output to remollout.root ... done
Graphics systems deleted.
Visualization Manager deleting...
 Running time[s]: 1233.91


I don't see any significant degradation of running time...
I am open to discuss about implementing it better way..
But I have many other pending jobs.. If someone agreed volunteer, then I can give the details of the geometry.

wdconinc: I don't know if it was ok previously. When I last checked (cubic interpolation) we spent most of the simulation time dealing with polycones. Not sure if those are collimators.

wdconinc (approved):

PR #490, opened 2021-05-08
Changing the length of the upstream outer nose shield from 100 mm to …

rahmans1:
Originally the length of the outer nose shield was 100 mm. Reducing it to 75.569 mm to avoid overlap with the geometric mother volume of the coils. The shield geometry in the upstream coil region for plates and nose shields are currently under optimization. So, this may be changed again in the near future.


wdconinc (approved):
Tested with checkoverlap, no issues. Thanks for fixing.

PR #501, opened 2021-05-27
[remoll] warn and exit if old geant4 or root version

wdconinc:
A mechanism to force updates...

Geant4 >= 4.10.6
ROOT >= 6.14.04


wdconinc:
I'm tired of supporting ancient geant4 versions :-)

rahmans1 (approved):
Sound good to me.

chandabindu:
We should stop using older version..

chandabindu (approved):

cameronc137 (approved):

PR #502, opened 2021-05-29
[searchPath] Fixed remoll not finding files

Yug34:


wdconinc (changes requested):
We can't just add /jlab/remoll. We have to figure out why this doesn't work. This way is just hiding the problem in one case.

wdconinc (approved):

PR #505, opened 2021-06-23
Bugfix: Use real momentum for physics generators instead of true momentum

rahmans1:
1) Flipped the sign of the axis to perform multiple scattering and other rotations. Effect negligible at shallow angles. But the problem shows up when using unidirectional large angle beams with physics generators. This commit should mitigate that.

2) Previously primary generator action used the true momentum instead of the real momentum for particle generation. Resulting in forward biased momentum all the time. Again negligible at shallow angles when beam is mostly parallel to the beamline but shows up for large angle beams.

Rate  in Ring 5 (900-1060 mm) before change:      122.32 GHz, 54.69 GHz, 45.94 GHz, 21.68 GHz
Rate in Ring 5 (900-1060 mm) after change:          121.97 GHz, 54.38 GHz, 45.93 GHz, 21.66 GHz

### DIFF
--- a/src/remollPrimaryGeneratorAction.cc
+++ b/src/remollPrimaryGeneratorAction.cc
@@ -172,24 +172,24 @@ void remollPrimaryGeneratorAction::GeneratePrimaries(G4Event* anEvent)
       fEvent = fEventGen->GenerateEvent();
       for (unsigned int pidx = 0; pidx < fEvent->fPartType.size(); pidx++) {
 
-        double p = fEvent->fPartMom[pidx].mag();
+        double p = fEvent->fPartRealMom[pidx].mag();
         double m = fEvent->fPartType[pidx]->GetPDGMass();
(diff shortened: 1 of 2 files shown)


rahmans1:
@wdconinc  Could you also take a look at this part of the code in remollVEventGen? The offset added to the vertex position seems to be with respect to origin at hall center. I think it should be with respect to origin at the beam origin. It's all shallow angles so the effect is probably negligible. Let me know what you think. Might need to translate the coordinate center to the beam origin -19.81 m, apply the position offset, translate back to hall coordinates. 
```
// Rotate position offsets due to multiple scattering
        for( iter = ev->fPartPos.begin(); iter != ev->fPartPos.end(); iter++ ) {
            //  rotate direction vectors based on multiple scattering
            (*iter) *= msrot;
        }
```

wdconinc (approved):

PR #518, opened 2021-09-14
Update showermax

sudipbhattarai:
Updated showermax position.
Added showermax virtual plane.
Updated showermax geometry with changed radii of upstream(open,close) and downstream(transition) rings.

### DIFF
--- a/geometry/mollerParallel.gdml
+++ b/geometry/mollerParallel.gdml
@@ -46,9 +46,12 @@
     startphi="0" deltaphi="360" aunit="deg"
     rmin="1450" rmax="1460" z="2700" lunit="mm"/>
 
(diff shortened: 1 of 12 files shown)


geometry/mollerParallel.gdml
```diff
@@ -46,9 +46,12 @@
     startphi="0" deltaphi="360" aunit="deg"
     rmin="1450" rmax="1460" z="2700" lunit="mm"/>
 
+  <tube
```
wdconinc, line 49: Not technically needed. You can just use ANY disk that already exists.

wdconinc (approved):

wdconinc (approved):

PR #525, opened 2021-09-28
changed showermax with stepped stack of tungsten and quartz

sudipbhattarai:


sudipbhattarai:
The idea of having a stepped stack of tungsten quartz is dropped.

PR #537, opened 2021-10-27
This will add scripts that I have made for the Punch Through Studies.

ktevans:
This includes punch through the sieve and blocker.

### DIFF
--- a/scripts/PunchThroughStudies/BlockerPunchThru.C
+++ b/scripts/PunchThroughStudies/BlockerPunchThru.C
@@ -1,5 +1,5 @@
 // Simple example macro for reroot analysis of remoll simulations for MOLLER
-//  
+//
 //  Need to use reroot (CERN's root compiled with special remoll libraries,
 //  or root when you have linked to ibremoll.so, i.e. by a command like
 //    setenv LD_PRELOAD build/libremoll.so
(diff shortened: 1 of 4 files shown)


wdconinc:
Can you rename the files so they don't have `#` in the name? That's a common comment character and may mess with people's ability to use the scripts.

scripts/PunchThroughStudies/BlockerPunchThru.C
```diff
@@ -0,0 +1,216 @@
...
+//  through the events and flag any in which a primary particle hits a Blocker hole.
+//  Then it will exclude these flagged events and plot electrons and positrons that hit
+//  upstream and downstream of the surface of the Blocker. Those particles that hit
+//  downstream of the surface are flagged as "punch through particles". 
+//  Punch-through particles are plotted on the rings of the main detector. An
+//  xy plot and radial plot are included. Finally, all charged particles that reach the
+//  main detector rings are plotted (xy and r).
+
+#include <TF1.h>
+void BlockerPunchThru##0()
```
wdconinc: This is the only one that will cause issues. The others, as far as I can tell, are fine.

wdconinc (changes requested):

PR #548, opened 2022-01-24
Feature shield design

zdemirog:
Shielding updates for config22.


cipriangal (approved):
This removes the large number of overlaps and puts in a new configuration.

PR #549, opened 2022-02-02
Change definition of fZOffset for external generator to be translation along Z-axis.

rahmans1:
…n of hits along z-axis instead of projection along direction of motion.


cipriangal (approved):
Looks good. Thank you for putting this back the way it was.

PR #551, opened 2022-02-04
Feature shield design

zdemirog:
Updates for the shielding configuration 23.


cipriangal (approved):
closely follows what is in the current engineering CAD model

PR #554, opened 2022-03-24
Optimized geometry with the new design of lead.

Elhamm1:
Optimized geometry with the new design of lead consists of 16*20cm Lead, 2 inches inner and outer shielding, 28 Lucite detectors, 1 PMT per detector, 23cm Lucite width

### DIFF
--- a/geometry/donut/donutConcreteLead.gdml
+++ b/geometry/donut/donutConcreteLead.gdml
@@ -126,7 +126,7 @@
     PMT radially outward; -90.0 degrees is with the PMT radially inward.
   -->
   <quantity name="donutPionDetectorSectorAngleBase" type="angle" value="+90.0" unit="deg"/>
-  <quantity name="donutPionDetectorSectorAnglePitch" type="angle" value="-87.1.0" unit="deg"/>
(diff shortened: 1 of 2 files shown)


geometry/donut/donutConcreteLead.gdml
```diff
@@ -126,14 +126,14 @@
     PMT radially outward; -90.0 degrees is with the PMT radially inward.
   -->
   <quantity name="donutPionDetectorSectorAngleBase" type="angle" value="+90.0" unit="deg"/>
-  <quantity name="donutPionDetectorSectorAnglePitch" type="angle" value="-90.0" unit="deg"/>
+  <quantity name="donutPionDetectorSectorAnglePitch" type="angle" value="-87.1.0" unit="deg"/>
```
wdconinc: This is not a number ;-)

wdconinc (changes requested):

geometry/donut/donutConcreteLead.gdml
Elhamm1: The typo was corrected.

PR #559, opened 2022-05-27
Changed the position of the virtual main detector

rahmans1:
Got new values from the CAD file


rahmans1:
Done as a demo for workshop. Not a real change. Closing without merging.

PR #563, opened 2022-08-03
Modify collimator 6 A

rahmans1:
**Work in Progress. Do not Merge Yet** Extracted the collimator 6A dimensions and positions from CAD file provided by JLAB (email chain with Daniel on 28 April, 2022). The details look different from what had previously in develop. May be a byproduct of shifting clamps. 

Once the https://github.com/JeffersonLab/remoll/pull/562 is merged, this PR needs to be merged to avoid collimator 6A overlap with new support structure.

Two keys points:
1) Given it's at a new z position, we probably need a new physics sculpt and propagate that through to JLAB CAD to avoid discrepancy. 
2) In CAD, the inner radial face of collimator 6A is a flat plane while we model it as a conical curved surface in simulation implementation. Does this matter?

See pictures below for updated collimator view:
Collimator6New

New

### DIFF
--- a/geometry/hybrid/hybridDaughter_unified.gdml
+++ b/geometry/hybrid/hybridDaughter_unified.gdml
@@ -555,19 +555,20 @@
         
         <!--    Solid for Collimator 6A -->
         <xtru name="Coll6A_solid1" lunit="mm">
-            <twoDimVertex x="52.77" y="-7.79"/>
-            <twoDimVertex x="46.90" y="-25.41"/>
-            <twoDimVertex x="91.58" y="-46.92"/>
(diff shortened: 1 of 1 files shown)


rahmans1:
Some pictures from the JLAB CAD for reference
6A
Collimator6A1

chandabindu:
I have checked the requirements for IR and OR for collimator 6A based on page 3 of docDB 969.

But, I feel (as shown in the attached pictures) that there are placement errors both radially and along z-direction.

Coll6A_Offsets
collimator6A

chandabindu (approved):

PR #566, opened 2022-08-07
Feature modify collimator5

rahmans1:
**Work in progress. Do not merge yet**. This should in principle resolve all overlap issues in the hybrid region with the integration of updated values from latest CAD. 

image
image

Running some test sims now to compare particle distribution at main detector plane with develop. Should be ready to merge once that's complete.

### DIFF
--- a/geometry/hybrid/hybridDaughter_unified.gdml
+++ b/geometry/hybrid/hybridDaughter_unified.gdml
@@ -443,7 +443,7 @@
         <subtraction name="Coll5_solid1">
             <first ref="Coll5_solid1_1"/>
             <second ref="Coll5_Cylinder_solid1"/>
-            <position name="Coll5_Sub_pos1" unit="mm" x="0" y="0" z="0" />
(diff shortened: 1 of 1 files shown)


chandabindu:
The implementation of collimator 5 seems okay based on page 7 of docDB 797.
But, the collimator 5 is moved upstream by 40.106 mm (Upstream face from z=7547.93 mm tp z=7507.824 mm.

The purpose of the 'edge-cut' at the inner side of collimator 5 is unclear to me. I believe it is to machining - to avoid knife-edge. Other aspects are okay.
We may have to come up with a new acceptance boundary for the inner edges of collimator 5.
But, that would be part of a future pull request. I'm okay with this pull request.
Coll5_Aug14

chandabindu (approved):

chandabindu:
I think we have missed something here.
Based on Dave Kashy's z-spreadsheet https://docs.google.com/spreadsheets/d/16ggbior7U50ckKhrI4Vr82Ux1w2bEm5Q/edit#gid=382248237 the upstream face of collimator 5 is 7558 mm. But, the current implementation is showing that is not correct.
collimator5

PR #569, opened 2022-08-21
Feature: update detector and SAM pipe region

rahmans1:
Updated the detector and SAM pipe region in accordance with the latest CAD file received from Randy Wilson (JLAB). 

Main changes:

1) Introduced material definitions for Inconel625 and Inconel625(10%)-Air(90%) mixture
2) Updated radial dimensions and z-extents of detector pipe, neckdown, SAM pipe, bellows and flanges. Set bellows flanges (yellow) to Inconel 625 material and the bellows to Inconel-Air mixture(brown). Set everything in blue to Aluminum including detector pipe, neckdown and SAM pipe with flange. 
ExperimentalGeometry
ZoomedExperimentalGeometry
3) Subtracted out a hole in hall logical volume and hall wall detector logical volume to avoid overlaps. 
4) Added an annulus of Aluminum as placeholders for the tubes holding the SAMs in place.
5) Moved the SAM placeholders to new location as per the CAD
6) Added sensitive detector ids for elements in the region after the neckdown (800-806)

### DIFF
--- a/geometry/beampipe/downstream/beampipeDSMother.gdml
+++ b/geometry/beampipe/downstream/beampipeDSMother.gdml
@@ -140,12 +140,16 @@
     <volume name="logic_Neckdown">
       <materialref ref="G4_Al"/>
       <solidref ref="Neckdown"/>
+      <auxiliary auxtype="SensDet" auxvalue="coilDet"/>
+      <auxiliary auxtype="DetNo" auxvalue="800"/>
       <auxiliary auxtype="Color" auxvalue="Blue"/>
     </volume>
 
     <volume name="logic_Neckdown_DSpipe">
       <materialref ref="G4_Al"/>
       <solidref ref="Neckdown_DSpipe"/>
+      <auxiliary auxtype="SensDet" auxvalue="coilDet"/>
+      <auxiliary auxtype="DetNo" auxvalue="801"/>
       <auxiliary auxtype="Color" auxvalue="Blue"/>
     </volume>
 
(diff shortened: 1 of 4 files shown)


cipriangal:
@rahmans1 I thought we didn't use the hall detectors in the actual geometry anymore. Could you remove them? (it will simplify the file and our lives). Once you do that let me know and I'll pull the geometry locally to have a closer look. From the PR here it looks great.

rahmans1:
@cipriangal These are the three virtual detectors (one for the cylindrical wall and two as lids) that exist in hallDaugter_dump.gdml. Do you want me to take them all out? Should they be removed completely or migrated to mollerParallel?
VirtualDet

cipriangal:
good points .. the round top can be removed (we never used it and it's overly complicated). The flat lid and the hall wall (which should be placed just shy of the wall itself) should be migrated to the parallel world (unless they are present there already). Thanks for checking this!

chandabindu:
Hi Sakib, 
I admit I haven't checked every detail.
But it looks okay visually. And it doesn't have any overlap issues.

the Al ring near the SAM inserts is ~57 mm thick and 177 mm long - Is it designed that way - it seems too much Al for me. Maybe it is required to support the SAMs.
The 'white' region near the hall end is Galactic. Could you place a cylinder with an approximate IR of SAM_Support_placeholder (=260.03 mm) of air to mimic the experimental condition? 
Thanks.

chandabindu (approved):
Please see the other comment.

cipriangal:
@rahmans1 : once you put the detectors in the moller parallel world i'll merge this. Thanks!

rahmans1:
> Hi Sakib, I admit I haven't checked every detail. But it looks okay visually. And it doesn't have any overlap issues.
> 
> the Al ring near the SAM inserts is ~57 mm thick and 177 mm long - Is it designed that way - it seems too much Al for me. Maybe it is required to support the SAMs. The 'white' region near the hall end is Galactic. Could you place a cylinder with an approximate IR of SAM_Support_placeholder (=260.03 mm) of air to mimic the experimental condition? Thanks.

@chandabindu  The Al ring is a placeholder. It will be replaced later by realistic tubes that insert into the SAM pipe. Both the z and azimuthal extent is the worst case possible so that we can have conservative estimates for splashback. The area around the SAM support holder should be air and inside it should be 1e-2 Torr (should be the case all along the beamline but I wanted to get all elements aligned with CAD before I fix the air/vacuum boundaries in a separate PR).

rahmans1:
> @rahmans1 : once you put the detectors in the moller parallel world i'll merge this. Thanks!

@cipriangal I have done the migration. Please check if they look right.

cipriangal:
everything looks good

PR #570, opened 2022-09-06
Increasing the azimuthal extent of the subtracted annulus for Collima…

rahmans1:
The inner radius of collimator 6B downstream piece wasn't properly sculpted because of incorrect azimuthal extent in subtracted annulus. This change should fix that.

### DIFF
--- a/geometry/hybrid/hybridDaughter_unified.gdml
+++ b/geometry/hybrid/hybridDaughter_unified.gdml
@@ -601,7 +601,7 @@
             <section zOrder="1" zPosition="0" xOffset="0" yOffset="0" scalingFactor="1"/>
             <section zOrder="2" zPosition="69.850" xOffset="0" yOffset="0" scalingFactor="1"/>
         </xtru>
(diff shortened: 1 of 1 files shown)


rahmans1:
@chandabindu A comparison between before and after. You can see narrow lines at the azimuthal extremes on the inner radius in the before case but not the after. Please double check before merging.
Before:
before
After:
Screenshot from 2022-09-06 14-26-50

I am wondering if there is any downside to using full azimuth annulus (startphi=0, deltaphi=360) for the subtracted volumes. That would remove the possibility of needing to tune this extent if other parameters change. @wdconinc Would there be any noticeable performance effect if a full azimuth cone is used for subtraction?

chandabindu (approved):
It looks good. The inner-edge radial cut looks good now for the collimator 6B.

chandabindu:
@wdconinc , could you please review and merge these changes to the develop branch? 
I am planning to run a few simulations with the updated changes. Thanks.

PR #573, opened 2023-01-16
Made a change in the documentation with instruction for users

rahmans1:


cipriangal:
Test for training a new student; doesn't need a pull

PR #574, opened 2023-02-02
TM2 and TM3 radial move-new default maps

rahmans1:
The subcoils in TM2 and TM3 torus magnets in the downstream spectrometer is now shifted outward by radius 3mm to allow more room for belly plate shielding and better clearance with central beam.  The new default field  map is subcoil_2_3_3mm_full.txt and the realistic asymmetric map for this case is subcoil_2_3_3mm_real_asymmetric.txt. Also adding the realistic asymmetric map for the TM0 torus magnet in upstream region.

Didn't notice any significant deterioration in deconvolution with new default maps.
https://moller.jlab.org/DocDB/0009/000981/001/OffsetCoil2and3.pdf

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -373,9 +373,12 @@ endif()
 # for i in *.txt ; do echo `md5sum $i|awk '{print$1}'`:`md5sum $i.gz|awk '{print$1}'`:$i ; done
 set(REMOLL_MAPS
 b8b30b803bae42f69135323ff40fe7db:99bf3fae83677c6c2f3e09a871682b2e:V2U.1a.50cm.parallel.txt
-d5f568675ca29e295c1013617a2dc092:b3e031ce9d97fa5331ec39e8903817e6:V2DSg.9.75cm.parallel.txt
+e4a96c2f3136d0eb020d90e74a532a6d:246d48001dd71174b62e16ecf07adc4d:V2U.1a.50cm.parallel.real_asymmetric.txt
+3ca8b6da4186d5de2a25bd89ef067a54:6a5fad69edd8cba457d85948bc42596b:subcoil_2_3_3mm_full.txt
+e6ab87da79a1dab8e759fcae6356e6f8:81b24e52bffbca83b3b7ca37b056f404:subcoil_2_3_3mm_real_asymmetric.txt
 )
 set(REMOLL_MAPS_EXTRA
+d5f568675ca29e295c1013617a2dc092:b3e031ce9d97fa5331ec39e8903817e6:V2DSg.9.75cm.parallel.txt
 1c79bc83e4da5e9d5e44338e28ce6ebe:c7591d167ca56e871e8e5eeecab2872c:V2DSg.9.75cm.txt
 f768447f761b9bf01af7c57e249d1d8a:94ec0017b21fc56d9d8084d68e479fd5:V2U.1a.50cm.txt
 b1b8f2845dc5cd420c6ff49ae7760a60:6002f6f5f092346cb2bdedac046d1905:V1U.2a.txt


cipriangal:
@rahmans1 which coils were offset? were the old ones in the wrong place and the b-field produced by TOSCA had a mistake? does this affect the actual coil geometry we have in the gdml?

rahmans1:
> @rahmans1 which coils were offset? were the old ones in the wrong place and the b-field produced by TOSCA had a mistake? does this affect the actual coil geometry we have in the gdml?

There was no mistake. Subcoils in segment 2 and 3 (TM2 and TM3) in the downstream spectrometer region have been consciously moved outward by 3 mm to allow more room for belly plate shielding if needed. This field map corresponds to this new default configuration. The physical positioning of the coils will be updated in a separate PR. Shouldn't be a noticeable effect as the move is negligible to the coil cross-section. 
image

cipriangal:
@rahmans1 thanks for clarifying.

Does this mean that collimator 5 needs to be moved as well?

From the overall results of the deconvolution you ran, I'd like to take a look at the big spreadsheet we usually make. Could you point me to that? The ~1sigma in some of the asymmetries is a bit concerning. We may have to live with it, but I just want to understand it.

rahmans1:
> @rahmans1 thanks for clarifying.
> 
> Does this mean that collimator 5 needs to be moved as well?
> 
> From the overall results of the deconvolution you ran, I'd like to take a look at the big spreadsheet we usually make. Could you point me to that? The ~1sigma in some of the asymmetries is a bit concerning. We may have to live with it, but I just want to understand it.

Collimator 5 stays where it is in the global coordinate system. Just the relative position of that element with the coils change since the coils are being moved outward. I am running an a version of deconvolution code that i had from a while ago. So, it would be good if @zdemirog double checked my numbers too. Here is the detailed table for the  [new default map](https://docs.google.com/spreadsheets/d/1SWnmVE_O8TvJkT-8IdretPhEZKELVVKooUbJQPmBNBA/edit#gid=0)

zdemirog:
> > @rahmans1 thanks for clarifying.
> > Does this mean that collimator 5 needs to be moved as well?
> > From the overall results of the deconvolution you ran, I'd like to take a look at the big spreadsheet we usually make. Could you point me to that? The ~1sigma in some of the asymmetries is a bit concerning. We may have to live with it, but I just want to understand it.
> 
> Collimator 5 stays where it is in the global coordinate system. Just the relative position of that element with the coils change since the coils are being moved outward. I am running an a version of deconvolution code that i had from a while ago. So, it would be good if @zdemirog double checked my numbers too. Here is the detailed table for the [new default map](https://docs.google.com/spreadsheets/d/1SWnmVE_O8TvJkT-8IdretPhEZKELVVKooUbJQPmBNBA/edit#gid=0)

I ran the simulation for the shielding configuration 36 by using new default field maps. You can find the results of the deconvolution study (along with ratio plots for the rate weighted radial distribution) [here](https://drive.google.com/file/d/1hHQklru8wwrfwkYbsI4cOtEp6GrJ-2gP/view?usp=sharing). The link of the detailed tables can be found on page 5.

cipriangal:
@rahmans1 : I am still looking at the deconvolution output (do you by any chance have a similar table like Zuhal pointed us to?).

However, it seems that if we move TM2 and TM3 we should see some impact on coll6a/b with asymmetric maps (https://moller.jlab.org/DocDB/0009/000974/004/NewCollimatorDesign.pdf). Did you look at this?

cipriangal:
@rahmans1 : Sorry about the table. Zuhal pointed out that you already sent it to me and I compared it to some recent results.

The first thing to notice is that the e-Al elastic is very different for Ring 2 closed:
Ring 2C	A	f
Sakib	74.3459	0.059782
Zuhal	985	0.00245102
Do you have any idea why this is?

The other thing to note is that in terms of f*A there are some big >10% differences for epElastic in ring 4C and epInelasticW1 for Ring 1O, Ring 4C. 

How about we discuss this at the detector meeting on Th? I am pretty sure these won't be too big of an issue but we should discuss them so we understand things better. 

If you manage to put together coll6a/b plots that would be something we can discuss carefully in the sim meeting as well.

rahmans1:
> @rahmans1 : Sorry about the table. Zuhal pointed out that you already sent it to me and I compared it to some recent results.
> 
> The first thing to notice is that the e-Al elastic is very different for Ring 2 closed: Ring 2C A f Sakib 74.3459 0.059782 Zuhal 985 0.00245102 Do you have any idea why this is?
> 
> The other thing to note is that in terms of f*A there are some big >10% differences for epElastic in ring 4C and epInelasticW1 for Ring 1O, Ring 4C.
> 
> How about we discuss this at the detector meeting on Th? I am pretty sure these won't be too big of an issue but we should discuss them so we understand things better.
> 
> If you manage to put together coll6a/b plots that would be something we can discuss carefully in the sim meeting as well.

The geometry I used to run my numbers are https://github.com/JeffersonLab/remoll/commits/develop head 83e4f37. Since then there have been multiple shielding updates. Some of the difference may be down to that. To truly compare apples to apples, I would probably need to rerun it with the latest develop. I am ok with discussing this on Thursday.

PR #575, opened 2023-02-07
Lead donut turned on in moller mother

rahmans1:
Seems like lead donut was turned off in moller mother in one of the shielding related PRs. Guessing this should be turned on.

### DIFF
--- a/geometry/mollerMother.gdml
+++ b/geometry/mollerMother.gdml
@@ -384,10 +384,10 @@
       </physvol>
       
       <!-- Lead Donut -->
-     <!-- <physvol>
+      <physvol>
 	<file name="donut/donutConcreteLead.gdml"/>
 	<positionref ref="donutConcreteLeadFaceUS"/>
-      </physvol>-->
+      </physvol>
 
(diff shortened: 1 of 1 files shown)


wdconinc:
I think the lead donut is now integrated through the pion system because they are not easily disentangled. @Elhamm1 can confirm.

rahmans1:
> I think the lead donut is now integrated through the pion system because they are not easily disentangled. @Elhamm1 can confirm.

Right now nothing relevant to pion detector is turned on in mollerMother. I am going by @Elhamm1 's branch where she has the donut system turned on https://github.com/Elhamm1/remoll/blob/PionDetectorSystemRotated/geometry/mollerMother.gdml. So, I am assuming that volume contains  everything relevant to pion.

wdconinc:
Yes, the pion was disabled in https://github.com/JeffersonLab/remoll/pull/571/. This is the right fix. I'll approve.

wdconinc (approved):

wdconinc:
I won't worry about the failing checks I guess.

PR #576, opened 2023-02-13
Shielding configuration 37

zdemirog:
Changed the material of the target roof from concrete to barite (this change has been made since config36). Added new virtual planes in the downstream torus region. Also, this configuration is done by importing the new changes from develop-branch (updated files are geometry/mollerMother.gdml and [geometry/pion/Lucite/pionDetectorLucite.gdml] ).


cipriangal (approved):
At first glance this looks good. We'll run additional checks with this branch before merging into develop.

PR #583, opened 2023-03-07
fixed the geometry issue in the US-end of the drift pipe

zdemirog:
This is just to fix the geometry issue that we noticed in the US-end of the drift pipe.


zdemirog:
here's the screenshot of the updated geometry.
comparison

PR #594, opened 2023-04-10
Introduce downstream torus tungsten belly plates in simulation

rahmans1:
Introducing downstream bellyplates with the following specs. The belly plates are 3 mm thick radially, cover the azimuthal extent of the coils and hug the bottom surface. 
BellyPlates222
Bellyplates2222


cipriangal:
@rahmans1 : did you look at the secondary sources at the MD from this region with these new elements?

PR #595, opened 2023-04-11
Feature showermax optical

sudipbhattarai:
Optical properties of shower-max is added and enabled. 
PMT Cathode has SD number 7xx16, where xx ranges from 00 (beam right open sector) to 27, stepping clockwise.
The gdml name showermaxgen.gdml is changed to showerMaxDetectorSystem.gdml . This name was used from old showermax, so, basically the file is replaced by the newest showermax model.


sudipbhattarai:
Not ready to merge yet.

sudipbhattarai:
This PR is ready to merge now.

rakithab:
@rahmans1, is the test geometry falling due to the same issue you mentioned?

rahmans1:
> @rahmans1, is the test geometry falling due to the same issue you mentioned?

Don't think so. This seems to be pointing to a issue with the geometry. I can't see the log for the failure though. So, speculating here. Can you retrigger the workflow?

rakithab:
I could not find a way to retrigger the workflow, is that something the pull request owner can do?

rakithab:
@rahmans1 I checked more recent pull requests and some checks are failing due to using the File map_directory/V2U.1a.50cm.parallel.txt which is probably not in use anymore.

rahmans1:
> @rahmans1 I checked more recent pull requests and some checks are failing due to using the File map_directory/V2U.1a.50cm.parallel.txt which is probably not in use anymore.

That is the most recent mapfile for the upstream though. Can you point me to the error log?

PR #605, opened 2023-08-13
Draft(DO NOT MERGE): Fix docker-related github actions failure

rahmans1:
Not sure what this line is for. But it's currently causing the github action to fail.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,6 @@
 FROM jeffersonlab/remoll-builder:main
 
 # XrootD
-RUN yum --enablerepo=extras -y install epel-release
 RUN yum -y install python3-pip \
     xrootd-client
 


rahmans1:
> Not sure what this line is for. But it's currently causing the github action to fail.

May be it's not related to that line but something in remoll-builder

wdconinc:
Probably just caused by the whole RHEL/Centos ecosystem's impending demise ;-)

rahmans1:
> Probably just caused by the whole RHEL/Centos ecosystem's impending demise ;-)

I guess the best course of action is to abandon it altogether?

wdconinc:
> I guess the best course of action is to abandon it altogether?

If it is not maintained, then yeah, I guess. But if someone wants to maintain it, then we can at least continue to run a minimum of quality control on pull requests. It probably just needs the remoll-builder to start from a non-redhat image. This is all built on at best 2-year old assumptions (probably more than that), so it needs a bit of review. That's the job of whoever is the simulation coordinator for MOLLER, I'd guess.

rakithab:
Just saw this, what actions are failing? Are those the ones with the red cross mark?

PR #610, opened 2024-02-23
Implements general rotation order

pranphy:
A new tag `<orotation>` is created that can be used in GDML files for ordered rotation.
```xml 
<orotation unit="degree" x="45" y="25" z="90" order="zxy"/>
````
this would allow the rotation 𝑅𝑧(90) → 𝑅𝑥(45) → 𝑅𝑦(25). The value of order parameter can be
any of the six permutations of string “xyz”. 

Example: 
image

▶ Yellow: Unrotated
▶ Green: 𝑅𝑥(90) → 𝑅𝑦(45) → 𝑅𝑧(0)
▶ Orange: 𝑅𝑦(45) → 𝑅𝑥(90) → 𝑅𝑧(0)
▶ there is a translation between them

Ref: [DocDB:1199](https://moller.jlab.org/cgi-bin/DocDB/private/ShowDocument?docid=1199)

### CODE REVIEW

**Comment by wdconinc** [comment]
FYI, this will not be valid GDML anymore. It will make schema validation always fail (unless you update the DTD in a derived schema and maintain that too). No other tools that interpret GDML (such are ROOT, FreeCAD) will be able to be used again. Be sure you know what you're doing when you merge this.

**Comment by pranphy** [comment]
@wdconinc  Thank you for your comment.  Does remoll/Geant4 parser do the schema validation by default? If we only ever use the gdml for remoll, only within remoll then it should be ok right? 
Also do you have a suggestion on how to implement the altered rotation order without breaking standard gdml schema?

**Comment by wdconinc** [comment]
> Also do you have a suggestion on how to implement the altered rotation order without breaking standard gdml schema?

Any rotation can be written as a rotation around three orthogonal axes, so there is a conversion, but it's probably not what you want to implement:
```
// from https://github.com/JeffersonLab/remoll/pull/611/files#diff-09868a28f9dae1133da8e941378ccc7725148e0979ae16308a56bcb8e324e14bR143
rxz = ROOT::Math::RotationZ(51.41960449999*TMath::DegToRad())*ROOT::Math::RotationX(3.00110158279097*TMath::DegToRad())
rzyx = ROOT::Math::RotationZYX(rxz)
cout << "x=" << rzyx.Psi() << " y=" << rzyx.Theta() << " z=" << rzyx.Phi() << std::endl;
ROOT::Math::Cartesian3D<double> v(1,1,1)
vp = rxz * v
vp2 = rzyx * v
cout << "x=" << vp.X() << " " << vp2.X() << " y=" <<  vp.Y() << " " << vp2.Y() << " z=" <<  vp.Z() << " " << vp2.Z() << std::endl;
```
produces:
```
x=0.0326825 y=0.0409392 z=0.896772
x=-0.116122 -0.116122 y=1.37184 1.37184 z=1.05098 1.05098
```

Easier may be to put the x rotations in e.g. geometry/detector/ThinQuartz/DetectorArray/FFSeg1_R5-FF-Quartz_L.gdml, and only do the z rotation in geometry/detector/ThinQuartz/DetectorArray/DetectorArray.gdml. That keeps the geometry modular.

cipriangal:
The layout of the quartz tiles using this new o-rotation

Screenshot 2024-02-27 at 9 11 59 AM

cipriangal:
@pranphy : We should discuss Wouter's suggestion at the simulation meeting today.

PR #622, opened 2025-09-26
Update CMakeLists.txt

rakithab:
HTTP no longer supported for files hosted at JLab. HTTPS  support has been added to CMake configuration so that it can auto-download map files.


rakithab:
Fixed the CMakeLists.txt to support https

PR #624, opened 2025-10-08
Update the rotation angle for the main detector in the GDML

Zhang-Xiang-nb:
We want all tiles in the main detector tilt toward to the beam line with a small angle, but currently we are not rotating the tiles  in the correct order. We suppose to rotate the detector with its intended local axis to facing the interaction point, but in the GDML geometry the rotation sequence is fixed. So we change the angle for each axis to make the final rotation in sequence x-y-z the same as sequence z-y-x. Based on the plots I got, it is working properly.
[rotationproblem.pptx](https://github.com/user-attachments/files/22781740/rotationproblem.pptx)


Zhang-Xiang-nb:
I also include the script which can convert the current gdml file to correct version in the geometry directory

Issue #256, opened 2019-04-09
Hardcoded target material name in remollGenElastic


wdconinc:
### Environment: (where does this bug occur, have you tried other environments)
- branch (master for latest released): master
- revision (HEAD for most recent): HEAD
- OS or system: all
- Special ROOT or Geant4 versions? no

### Steps to reproduce: (give a step by step account of how to trigger the bug)
https://github.com/JeffersonLab/remoll/blob/f626940f05915a3ba559fad296c551f94e9f0d9f/src/remollGenpElastic.cc#L52

### Expected Result: (what do you expect when you execute the steps above)
This should not depend on some target material.

### Actual Result: (what do you get when you execute the steps above)
It does.

Not sure what to do about it...

rahmans1:
@wdconinc  What are the implications of this? If you don't have a material named "LiquidHydrogen", the generator will malfunction?

wdconinc:
Potentially... the ep elastic generator will do something different.

wdconinc:
Some work towards this in 4be2ee8a60ddc537bae1e57b48ea238abf3e6519

wdconinc:
We should use something like:
```
	  const G4ElementVector* ev = (*it)->GetLogicalVolume()->GetMaterial()->GetElementVector();
	  for (auto e = ev->begin(); e != ev->end(); e++) {
	    G4cout << (*e)->GetZ() << G4endl;
	  }
```
This would at least make sure there is hydrogen in the target.

PR #630, opened 2025-11-12
updates on LH2 target properties: implementing realistic temperature, pressure, density values for LH2

zdemirog:
Update of LH2 material properties in remoll. 
Details at [docDB1502](https://moller-docdb.physics.sunysb.edu/cgi-bin/DocDBTest/private/ShowDocument?docid=1502)

### DIFF
--- a/geometry/materials.xml
+++ b/geometry/materials.xml
@@ -240,10 +240,17 @@
     <D unit="g/cm3" value="1e-25"/>
     <atom unit="g/mole" value="1.01"/>
   </material>
-  <material Z="1" name="LiquidHydrogen" state="liquid">
+  <!-- <material Z="1" name="LiquidHydrogen" state="liquid">
     <T unit="K" value="20.27"/>
     <D unit="g/cm3" value="0.0708"/>
     <atom unit="g/mole" value="1.00794"/>
+  </material>-->
+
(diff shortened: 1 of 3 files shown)


zdemirog:
On [c79ea74](https://github.com/JeffersonLab/remoll/pull/630/commits/c79ea74c5655d7b35d3b896f14ce93c861afdd5e), I updated the LiquidHydrogen material from a simple element definition to a composite material definition, so that Geant4 creates valid AtomsVector, ElementVector etc, which are required by the active target volume;


```
	    const G4int *atomvec = material->GetAtomsVector();
	    const G4ElementVector *elvec = material->GetElementVector();
	    const G4double *fracvec = material->GetFractionVector();
```
https://github.com/JeffersonLab/remoll/blob/develop/src/remollBeamTarget.cc#L369C1-L371C62

I also changed the hard coded [target material name](https://github.com/JeffersonLab/remoll/blob/develop/src/remollGenpElastic.cc#L47-L52) in ep-elastic generator (issue #256 ), so it finds and uses the new customized LiquidHydrogen material instead of G4_lH2 name.


Updated simulation results related to this update can be found in [docDB1502](https://moller-docdb.physics.sunysb.edu/cgi-bin/DocDBTest/private/ShowDocument?docid=1502)

PR #636, opened 2025-12-07
the trigger scintillator has been added to the moller setup at the pr…

afezland89:
This is a updated copy of our trigger scintillator to the most recent version of the remoll project repo.


rakithab:
No longer needed

PR #642, opened 2026-05-12
Feature add sbms

dvalmassei:
Adding LAMs.
-sensDets: 8XXYY; where XX represents the sector (00-06), and YY represents radiator (01), left PMT (02), or right PMT (03)


dvalmassei:
This PR is ready to merge

PR #643, opened 2026-05-31
Add Showermax stackplanes and update response lookup 

sudipbhattarai:
The Shower-max stackplanes (28 virtual planes positioned in front of individual Shower-max modules) are placed. These detectors are the standard planes used in most Shower-max-related remoll studies. The SD numbers for these planes range from 73000 to 73028, starting from the beam left closed sector and progressing clockwise.

The lookup table data and script have been reorganized. Anyone interested in using this table can simply add one header file and use the GetResponse function.


sudipbhattarai:
This PR is ready to merge.

